PR #32, opened 2015-01-26
Build fixes for OS X and latest rust

mwylde:
Also cleans up some warnings from the int/uint to isize/usize migration.


mrmonday:
Thank you very much!

PR #36, opened 2015-01-28
Netmap support

mrmonday:
This pull requests adds experimental support for netmap. Basic testing seems to work, though test cases do not currently pass. It is likely that there are serious bugs in the `unsafe` code, but they haven't shown themselves yet. Compile with `--features netmap` to test it.


mrmonday:
I should also note that this only currently works on Linux - FreeBSD support is currently missing.

PR #37, opened 2015-01-30
More rust fixes

mrmonday:


mrmonday:
Still warnings with netmap, can't seem to stop that though... Doesn't matter too much since this is just a temporary hack.

PR #41, opened 2015-02-05
Fix Windows build

cybergeek94:
Fixes:
* Windows build
* Win64 segfaults

Does not fix:
* Win32 hanging (Win64 now hangs)

Still needs done:
* `.cargo/config` needs the target configuration duplicated for the Win32 target triple that I can't remember


mrmonday:
Thank you so much for your work!

brson:
Nice teamwork! :+1:

PR #49, opened 2015-02-20
Fixes for latest Rust master.

mrmonday:


mrmonday:
Gonna merge this, despite OS X being broken - it would seem the nightlies are using different revisions currently, so it's impossible to have it working on both Linux and OS X. Since I'm developing on Linux, OS X will have to fall behind for now.

PR #56, opened 2015-03-26
Fixing bug in `get_network_interfaces_impl` interface merging.

### LINKED ISSUE #55: Issue getting IPs for interfaces on OSX
I'm having an issue getting IP addresses for interfaces on OSX using `get_network_interfaces`.

The problem appears to be this code: https://github.com/libpnet/libpnet/blob/ea45c1d8b7d91b98ebe4b873d2f77fcd71c632dd/src/util.rs#L226-236.

If `old.ips` is `None`, (i.e. the first interface we found for a given name has no IPs), then the new ones will never be merged in. 

I know you're working on a rewrite, so I didn't bother opening a PR, but would be happy to do so (with an attempt at a test demoing the issue) if you'd like.

lukeasrodgers:
If we found an interface, but it had no IPs, subsequent interfaces of the same name with IPs would not be merged into the interface struct.

Adds a test for non-linux platforms, following the example of the `check_test_environment` test.

@mrmonday let me know if you'd like an additional test for alternate scenarios (e.g. in which no IP addresses should be found for a given interface). Not sure exactly how I'd implement that yet, might require some refactoring of `get_network_interfaces_impl` but I could give it a shot. Also not sure how this change will fare on non-OSX platforms.

Fixes #55

### DIFF
--- a/src/test.rs
+++ b/src/test.rs
@@ -22,6 +22,7 @@ use old_packet::udp::{UdpHeader, MutableUdpHeader, UdpPacket};
 use transport::{udp_header_iter, ipv4_header_iter, transport_channel, TransportProtocol,
                 TransportChannelType};
 use transport::TransportProtocol::{Ipv4, Ipv6};
+use util;
 
 const IPV4_HEADER_LEN: usize = 20;
 const IPV6_HEADER_LEN: usize = 40;
@@ -234,33 +235,33 @@ fn layer3_ipv4() {
     res.join();
 }
 
+fn get_test_interface() -> util::NetworkInterface {
+    use std::clone::Clone;
+    use std::env;
+
+    (*util::get_network_interfaces()
+        .as_slice().iter()
+        .filter(|x| {
+            match env::var("PNET_TEST_IFACE") {
+                Ok(name) => x.name == name,
+                Err(_) => x.is_loopback()
+            }
+        })
+        .next()
+        .unwrap())
+        .clone()
+}
+
+
 // FIXME Find a way to test this with netmap
 #[cfg(not(feature = "netmap"))]
 #[test]
 fn layer2() {
     use datalink::{datalink_channel, DataLinkChannelType};
     use old_packet::ethernet::{EtherTypes, EthernetHeader, MutableEthernetHeader};
-    use util;
 
     const MIN_PACKET_SIZE: usize = 64;
     const ETHERNET_HEADER_LEN: usize = 14;
-
-    fn get_test_interface() -> util::NetworkInterface {
-        use std::clone::Clone;
-        use std::env;
-
-        (*util::get_network_interfaces()
-            .as_slice().iter()
-            .filter(|x| {
-                match env::var("PNET_TEST_IFACE") {
-                    Ok(name) => x.name == name,
-                    Err(_) => x.is_loopback()
-                }
-            })
-            .next()
-            .unwrap())
-            .clone()
-    }
 
     let interface = get_test_interface();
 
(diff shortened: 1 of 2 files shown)


src/util.rs
```diff
@@ -228,10 +228,15 @@ fn get_network_interfaces_impl() -> Vec<NetworkInterface> {
             None => old.mac,
             _ => new.mac
         };
-        match (&mut old.ips, &new.ips) {
-            (&mut Some(ref mut old_ips), &Some(ref new_ips)) => old_ips.push_all(new_ips.as_slice()),
-            _ => {}
-        };
+        if match (&old.ips, &new.ips) { (&None, &Some(_)) => true, _ => false } {
+            old.ips = new.ips.clone();
+        }
```
mrmonday: Could this not be merged into the `match` statement below?

src/test.rs
```diff
@@ -343,3 +344,20 @@ fn check_test_environment() {
     #[cfg(target_os = "linux")]
     fn test_iface() {}
 }
+
+#[test]
+fn test_get_network_interfaces_ips() {
+    test_iface_ips();
```
mrmonday: I'm not convinced that this is the correct way to test this - it's entirely possible that the selected test interface will not have an IP address assigned to it. This said, the whole testing mechanism is a bit of a hack currently, so perhaps this is OK...

src/util.rs
lukeasrodgers: Possibly, except since the first element in the match tuple will be `None` we won't have a mutable reference to `old_ips` to which we could assign `new_ips`, but I'm still learning rust so I could be missing something obvious.

src/test.rs
lukeasrodgers: Okay, yeah I wasn't sure if that was a reasonable expectation. I'll see if I can figure out another way to test this.

src/util.rs
mrmonday: You should be able to use `(old_ips @ &mut None, &Some(ref new_ips)) => *old_ips = Some(new_ips.as_slice())`, or something to that effect.

lukeasrodgers: Did not know you could do that, I'll try it out, thanks.

src/test.rs
lukeasrodgers: So, I agree that this is a suboptimal way to test the method, but it seems that a prerequisite for more comprehensive and reliable tests would involve a fairly significant rewrite of almost all the code in `get_network_interfaces_impl`. 

I'd be happy to attempt such a rewrite if you thought it advisable/worth it, but as I mentioned my rust-fu is weak, so I'd probably need a fair bit of feedback on my attempts to do so. If that's the route you'd like to go, we could maybe just get rid of the test for this PR, and then tackle that in another PR.

Open to other suggestions, etc.

mrmonday: I think the right way to test this is dependency injection, replacing `libc::getifaddrs()` then testing that `get_network_interfaces()` returns the correct set of `NetworkInterface`s. This is probably overkill for this fix though - if you want to work on it, I'd love a proper test suite for it - if not I'll happily accept the pull sans-test case (I don't think it's a good idea to include the current test case).

lukeasrodgers: Sounds good, that's roughly what I was thinking too. I'll take a stab at it.

mrmonday:
@lukeasrodgers, are you still working on this? I'd like to get the fix merged in.

lukeasrodgers:
@mrmonday sorry, am suffering from debilitating tendinitis in both thumbs and am avoiding typing. Probably won't be able to work on this for another month or so. Proceed as you see fit.

mrmonday:
I'm sorry to hear that. I'll try and get this cleaned up merged when I can.

mrmonday:
Opened a new pull request (pull #75) without the tests but with your fix.

PR #57, opened 2015-03-27
Fixes for latest Rust master.

mrmonday:


mrmonday:
Haven't figured out why the tests are failing yet, I'm hoping they'll magically fix themselves once the new packet stuff is merged.

mrmonday:
I should probably note that they're passing locally, so it could be something weird with Travis' setup.

PR #67, opened 2015-05-14
Add function to get the minimum packet size

vmx:
The `minimum_packet_size()` returns the total number of bytes the
fixed-size elements have.

With this function it's possble to determine the size of the packet buffer
that is needed to be populated by a struct. Only the variable sized fields
need to be added to the size.


mrmonday:
It would be nice to have test cases for this, but since the infrastructure isn't really set up for it yet I'll merge it once Travis reports back.

vmx:
I actually thought as well that there should really be a unit test for it. I'll have a look how hard it'll be to add one.

PR #71, opened 2015-05-24
Error when struct/enum isn't public

vmx:
The should be an error when a struct that has an `#[packet]` attribute isn't
public. This is an implementation for the "must_be_pub" test.

### DIFF
--- a/pnet_macros/src/decorator.rs
+++ b/pnet_macros/src/decorator.rs
@@ -213,6 +213,10 @@ fn make_packet(ecx: &mut ExtCtxt, span: Span, name: String, sd: &ast::StructDef)
 fn make_packets(ecx: &mut ExtCtxt, span: Span, item: &ast::Item) -> Option<Vec<Packet>> {
     match item.node {
         ast::ItemEnum(ref ed, ref _gs) => {
+            if item.vis != ast::Visibility::Public {
+                ecx.span_err(item.span, "#[packet] enums must be public");
+                return None;
+            }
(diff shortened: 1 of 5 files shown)


pnet_macros/src/decorator.rs
```diff
@@ -211,6 +211,10 @@ fn make_packet(ecx: &mut ExtCtxt, span: Span, name: String, sd: &ast::StructDef)
 }
 
 fn make_packets(ecx: &mut ExtCtxt, span: Span, item: &ast::Item) -> Option<Vec<Packet>> {
+    if item.vis != ast::Visibility::Public {
+        ecx.span_err(item.span, "#[packet] structs must be public");
```
mrmonday: Should be "structs and enums".

pnet_macros/src/decorator.rs
```diff
@@ -211,6 +211,10 @@ fn make_packet(ecx: &mut ExtCtxt, span: Span, name: String, sd: &ast::StructDef)
...
+        return None
```
mrmonday: Could you add a semi-colon at the end of this line please?

mrmonday:
How's your work on the pnet_macros tests coming along? Perhaps it would be a good idea to enable the test-suite for Travis soon?

pnet_macros/src/decorator.rs
mrmonday: I suppose you'll also need to update the test case with the relevant error message too.

PR #72, opened 2015-05-25
Fix libpnet_macros tests

vmx:
When all these commits are merged there will be the libpnet_macros tests running on the CI.


mrmonday:
This looks great, thank you! A couple of comments:

 * Just spotted this file - pnet_macros/tests/run-pass/ipv4.generated.rs_ - that should probably be deleted - it shouldn't be there, must have slipped in by mistake. Same for any other similarly named files.
 * The Makefile may as well be deleted for pnet_macros, since it can all be run through cargo now. Make sure to update build.sh so that it passes along $VERBOSE or $CARGO_FLAGS, or whatever the variable is in there (I forget).

mrmonday:
Could you also add a call to `run_macro_tests` in `travis_script()` in build.sh please?

mrmonday:
Great work! Thank you very much!

PR #74, opened 2015-05-28
Adding #[length]

vmx:
This commits add an alternative way to provide the length of a variable sized field. Instead of providing a function, an arithmetic expression like `#[length] = "field_name - 4"` is specified.

The advantage of this approach is that no custom functions are needed although it's probably still powerful enough for most cases.

In the future this expression can be used determine the size of the source struct even for variable sized fields.


mrmonday:
I don't have time to give this a proper code review right now, I'll take a look tomorrow though. This definitely looks very promising!

mrmonday:
This is great, thank you!

PR #77, opened 2015-05-29
Require a minimum length for packets.

mrmonday:
Constructing a new packet will now return an Option, which will be None if the
provided buffer is not sufficiently large.


mrmonday:
Whilst not currently done, this should allow using get_unchecked() for accessing the buffer in most cases, removing the overhead of bounds checking.

vmx:
That's a nice change.

PR #79, opened 2015-06-13
Add function to determine the packet_size from a struct

vmx:
There's a new `packet_size` function that is implemented for the
packets in case the size of a packet can be determined from the
struct. This is the case when the sizes of all variable sized fields
are defined by a length expression (as opposed to a length function).

This makes it possible to create a new packet from a struct without
knowing its size up front. Example:

    #[packet]
    pub struct Simple {
        sz: u8,
        #[length = "sz"]
        #[payload]
        payload: Vec<u8>
    }

    let payload = vec![1, 2, 3, 4];
    let simple = Simple {
        sz: payload.len() as u8,
        payload: payload
    };

    let mut packet = vec![0; SimplePacket::packet_size(&simple)];
    MutableDcpOpenPacket::new(&mut packet[..]).unwrap().populate(simple);

### DIFF
--- a/pnet_macros/src/decorator.rs
+++ b/pnet_macros/src/decorator.rs
@@ -47,6 +47,7 @@ struct Field {
     span: Span,
     ty: Type,
     packet_length: Option<String>,
+    struct_length: Option<String>,
     is_payload: bool,
     construct_with: Option<Vec<Type>>
 }
@@ -96,6 +97,7 @@ fn make_packet(ecx: &mut ExtCtxt, span: Span, name: String, sd: &ast::StructDef)
         };
         let mut is_payload = false;
         let mut packet_length = None;
+        let mut struct_length = None;
         let mut construct_with = Vec::new();
         let mut seen = Vec::new();
         for attr in field.node.attrs.iter() {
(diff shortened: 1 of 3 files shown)


mrmonday:
Great work, thanks again!

mrmonday:
This might actually be the most motivating example I've seen for getting rid of `#[length_fn]`.

I'm not ready to remove it yet, but if we build up a good base of `#[packet]`s in `pnet::packet` and it's not needed then I'd definitely be open to it.

vmx:
Thanks for merging. I just had another thought. Instead of having the function not defined if it's not possible, should it perhaps return an `Option` or `Result` instead?

mrmonday:
Just a thought actually - the `packet_size` function should probably always be defined - it should always be possible to figure out the size of the packet. In the test that's currently compile-fail, it should return `5`.

vmx:
Wow, it's pretty obvious that I could just do `len()` calls on the vectors of the variable sized fields. I introduced the whole length expression stuff more or less just to make the patch above happen. I should've thought about all this harder up front ;)

mrmonday:
It's still a useful change, far more concise than defining an extra function :)

PR #80, opened 2015-06-14
Remove information about `old_packet` from the README.

mrmonday:


mrmonday:
@homu r+

Let's see if I've got everything set up right for homu :)

PR #87, opened 2015-07-04
Fix packet range access

vmx:
When getting a slice via a range, the end of the range is not the relative
offset from the start of the range, but the absolute position counted from
the start of the original array.


mrmonday:
@homu r+

Good work, thanks.

PR #88, opened 2015-07-07
Add missing documentation.

mrmonday:
This adds in some documentation which Rust has recently started complaining about being missing. It's odd that it wasn't caught already.

Also switched from warn -> deny for missing documentation in pnet_macros, since it is completely documented.


mrmonday:
@homu r+

PR #94, opened 2015-07-31
Bump the version number for libpnet

mrmonday:
This should have been done long ago, since the version currently on crates.io
doesn't compile.


mrmonday:
@homu r+

PR #98, opened 2015-08-11
code review modular checksum functionality

david415:
This is a rough draft implementation of a modular checksum mechanism that has lots of code reuse for each protocol making use of the rfc1071 checksum. This dev branch has lots of intermediate commits that make a sloppy commit history... so we should probably just review the final diff instead of the individual commits.


AUTHORS
```diff
@@ -0,0 +1,6 @@
+
```
mrmonday: I see no need to add an AUTHORS file, that's what git is for.

src/packet/checksum.rs
```diff
@@ -0,0 +1,21 @@
+//! rfc1071 checksum abstraction
```
mrmonday: This file should have the same header as the other files. I also don't know that this is the right place for this code, perhaps `pnet::util` would be a better location.

src/packet/checksum.rs
```diff
@@ -0,0 +1,21 @@
...
+
+/// Calculates rfc1071 checksum value
+pub fn rfc1071_checksum(packet: &[u8], initial: u32) -> u16 {
```
mrmonday: There needs to be a unit test for this function.

src/packet/checksum.rs
```diff
@@ -0,0 +1,21 @@
...
+    let length = packet.len() - 1;
+    let mut sum = initial;
+
+    let mut i = 0;
+    while i < length {
+        let word = (packet[i] as u32) << 8 | packet[i + 1] as u32;
+        sum = sum + word;
+        i = i + 2;
+    }
+    if packet.len()%2 == 1 {
```
mrmonday: Please use spaces either side of binary operators.

src/packet/ipv4.rs
```diff
@@ -39,22 +41,43 @@ pub struct Ipv4 {
     payload: Vec<u8>,
 }
 
+
+impl <'p> HasPseudoheader for Ipv4Packet <'p> {
```
mrmonday: No spaces before lifetimes, eg `impl<'p>`.

src/packet/ipv4.rs
```diff
@@ -39,22 +41,43 @@ pub struct Ipv4 {
...
+    fn pseudoheader_checksum(&self) -> u32 {
```
mrmonday: I'd like to see some benchmarks for old-style vs new style. Theoretically they should be optimised to the same thing based on my quick look over, but I would rather be sure.

src/packet/mod.rs
```diff
@@ -34,6 +34,12 @@ pub trait MutablePacket : Packet {
     }
 }
 
+/// Used to help compute checksum value of packet
+pub trait HasPseudoheader {
```
mrmonday: I think a better name for this trait would be `PseudoHeader`.

src/packet/mod.rs
```diff
@@ -34,6 +34,12 @@ pub trait MutablePacket : Packet {
...
+    /// Converts IPv4 or IPv6 specific fields to a checksum value
+    fn pseudoheader_checksum(&self) -> u32;
```
mrmonday: Since the trait name contains pseudo-header, there's no need to prefix this method - simply call it checksum.

src/packet/mod.rs
```diff
@@ -78,9 +84,11 @@ impl PrimitiveValues for ::std::net::Ipv6Addr {
     }
 }
 
+pub mod checksum;
 pub mod ethernet;
 pub mod ip;
 pub mod ipv4;
 pub mod ipv6;
 pub mod udp;
+pub mod tcp;
```
mrmonday: I'd rather your work on TCP be separated into a separate pull request - I'll review it here anyway though.

src/packet/tcp.rs
```diff
@@ -0,0 +1,197 @@
...
+
+fn tcp_options_length<'a>(tcp: &TcpPacket<'a>) -> usize {
+    /* The data offset field specifies the total size of the TCP header
+       in 32-bit words. The minimum size of a TCP header is 20 bytes and thus
+       TCP headers must have data offset set to 5 or more.
+     */
+    let mut v = tcp.get_data_offset_and_reserved();
+    v  = v >> 4; // get rid of reserved bits
+    v = v - 5; // remove the size of the minimal tcp header of 20 bytes
+    v = v * 4; // multiple the remaining word count by 4 to get the byte count
```
mrmonday: Is this bit shifting and arithmetic really necessary? The whole reason for `#[packet]` to exist is to remove the need for manual shifting, which is incredibly error prone.

src/packet/tcp.rs
```diff
@@ -0,0 +1,197 @@
...
+//! TCP packet abstraction
+
+use packet::Packet;
+use packet::HasPseudoheader;
+use packet::checksum::rfc1071_checksum;
+use pnet_macros::types::*;
+
+/// Represents a TCP Packet
+#[packet]
+pub struct Tcp {
```
mrmonday: I haven't checked the structure of this packet to make sure it matches the spec - I'll do that in a later review.

src/packet/tcp.rs
```diff
@@ -0,0 +1,197 @@
...
+    let mut v = tcp.get_data_offset_and_reserved();
+    v  = v >> 4; // get rid of reserved bits
+    v = v - 5; // remove the size of the minimal tcp header of 20 bytes
+    v = v * 4; // multiple the remaining word count by 4 to get the byte count
+    return v as usize;
+}
+
+/// Represents the TCP Option fields
+#[packet]
+pub struct TcpOptions {
```
mrmonday: Does this represent a single option or multiple options? If the former, this should be named `TcpOption`.

src/packet/tcp.rs
```diff
@@ -0,0 +1,197 @@
...
+    padding: Vec<TcpPadding>,
+    #[payload]
+    payload: Vec<u8>
+}
+
+fn tcp_options_length<'a>(tcp: &TcpPacket<'a>) -> usize {
+    /* The data offset field specifies the total size of the TCP header
+       in 32-bit words. The minimum size of a TCP header is 20 bytes and thus
+       TCP headers must have data offset set to 5 or more.
+     */
```
mrmonday: Please use line comments rather than multi-line comments.

src/packet/tcp.rs
```diff
@@ -0,0 +1,197 @@
...
+    /* The TCP header padding is used to ensure that the entire header
+       ends on a 32 bit boundary.
+     */
+    //return tcp.get_data_offset_and_reserved() as usize % 4;
+    return 0
+}
+
+/// Represents the TCP header padding
+#[packet]
+pub struct TcpPadding {
```
mrmonday: If padding is simply a number of u8's, does this really need its own `#[packet]`?

src/packet/tcp.rs
```diff
@@ -0,0 +1,197 @@
...
+    #[payload]
+    data: Vec<u8>
+}
+
+fn tcp_padding_length<'a>(tcp: &TcpPacket<'a>) -> usize {
+    /* The TCP header padding is used to ensure that the entire header
+       ends on a 32 bit boundary.
+     */
+    //return tcp.get_data_offset_and_reserved() as usize % 4;
+    return 0
```
mrmonday: This obviously needs fixing :)

src/packet/tcp.rs
```diff
@@ -0,0 +1,197 @@
...
+/// Represents the TCP header padding
+#[packet]
+pub struct TcpPadding {
+    #[payload]
+    data: Vec<u8>
+}
+
+/// Calculates the checksum of a TCP packet
+/// The passed in TcpPacket must have it's initial checksum value set to zero.
+pub fn checksum<'a, T: HasPseudoheader>(packet: &TcpPacket<'a>, encapsulating_packet: T) -> u16be {
```
mrmonday: Does `encapsulating_packet` really need to take ownership? How well does this work? I've previously avoided having multiple packets existing for the same buffer for ownership reasons - it can become a hassle trying to work around it.

src/packet/mod.rs
```diff
@@ -78,9 +84,11 @@ impl PrimitiveValues for ::std::net::Ipv6Addr {
     }
 }
 
+pub mod checksum;
 pub mod ethernet;
 pub mod ip;
 pub mod ipv4;
 pub mod ipv6;
 pub mod udp;
```
mrmonday: The checksums in UDP will need to be converted to this new method of checksumming too.

src/packet/tcp.rs
```diff
@@ -0,0 +1,197 @@
...
+    return rfc1071_checksum(packet.packet(), sum);
+}
+
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    const TCP_MIN_HEADER_LEN: usize = 20;
+    const TCP_OPTIONS_LEN: usize = 12;
```
mrmonday: These two constants shouldn't be needed any more thanks to the `minimum_packet_size()` associated functions.

src/packet/tcp.rs
```diff
@@ -0,0 +1,197 @@
...
+        {
+            let mut mut_ip_header = MutableIpv4Packet::new(&mut packet[..]).unwrap();
+            mut_ip_header.set_next_level_protocol(next_level_protocol);
+            mut_ip_header.set_source(ipv4_source);
+            mut_ip_header.set_destination(ipv4_destination);
+        }
+
+        {
+            let mut mutable_tcp_header = MutableTcpPacket::new(&mut packet[IPV4_HEADER_LEN..]).unwrap();
+            mutable_tcp_header.set_checksum(0);
```
mrmonday: The packet is initialised to zeroes anyway - this line isn't necessary.

src/packet/tcp.rs
```diff
@@ -0,0 +1,197 @@
...
+
+        {
+            let ip_header = Ipv4Packet::new(&packet[..IPV4_HEADER_LEN]).unwrap();
+            let tcp_header = TcpPacket::new(&packet[IPV4_HEADER_LEN..]).unwrap();
+            csum = checksum(&tcp_header, ip_header);
+        }
+
+        {
+            let mut mutable_tcp_header = MutableTcpPacket::new(&mut packet[IPV4_HEADER_LEN..]).unwrap();
+            mutable_tcp_header.set_checksum(csum);
```
mrmonday: It's very unfortunate that setting a checksum is this verbose - it would be great if this could be improved.

src/packet/tcp.rs
```diff
@@ -0,0 +1,197 @@
...
+
+        tcp_header.set_checksum(0xfe2d);
+        assert_eq!(tcp_header.get_checksum(), 0xfe2d);
+
+        tcp_header.set_urgent_pointer(0x0000);
+        assert_eq!(tcp_header.get_urgent_pointer(), 0x0000);
+
+        let tcp_option = TcpOptions{
+            kind: 0x1,
+            data: vec![0x01,0x08,0x0a,0x1d,0xfc,0xcb,0x76,0x1d,0xfc,0xbe,0x62],
```
mrmonday: The `TcpOptions` `#[packet]` needs improving - there should be no need to deal with all this raw data. The same applies to some other fields too. The interface needs to be as high level as possible, without losing functionality.

src/packet/tcp.rs
david415: it might not be needed if we can specify struct attributes which represent 4 bits...

src/packet/tcp.rs
david415: how else to represent a variable number of u8's in the packet?

src/packet/tcp.rs
david415: oops... yeah i need to think about packet composition and what is really needed here

src/packet/mod.rs
david415: yes...

src/packet/tcp.rs
mrmonday: You're looking for `u4`.

src/packet/tcp.rs
mrmonday: You can just use `Vec<u8>`, along with a `#[length]` or `#[length_fn]`.

src/packet/tcp.rs
david415: Yes I agree... this was temporarily raw. I'll write a high level abstraction for specifying the options.

src/packet/tcp.rs
david415: we could encapsulate all of that into a method call for MutableTcpPacket...

mrmonday:
Hi,

I'm closing this pull request since it doesn't build and there has been no activity since last year. Please feel free to reopen it once it's building again!

PR #104, opened 2015-08-21
91.add modular checksum.1

david415:
Can I receive another code review? I believe I've made the improvements you've specified in our last code review. This dev branch doesn't include any TCP stuff... I am merely trying to apply some code reuse to the checksum functions for ipv4, ipv6 and udp... However right now I'm not that good at rust and I tend to fight with the compiler over multiple references in these unit tests which thus far I have not gotten to build yet.

### DIFF
--- a/src/packet/ipv4.rs
+++ b/src/packet/ipv4.rs
@@ -9,9 +9,9 @@
 //! IPv4 packet abstraction
 
 use packet::ip::IpNextHeaderProtocol;
-
+use packet::{Pseudoheader,PrimitiveValues};
+use util::rfc1071_checksum;
 use pnet_macros::types::*;
-
 use std::net::Ipv4Addr;
 
 /// Represents an IPv4 Packet
@@ -39,8 +39,8 @@ pub struct Ipv4 {
     payload: Vec<u8>,
 }
 
-impl <'p> HasPseudoheader for Ipv4Packet <'p> {
-    fn pseudoheader_checksum(&self) -> u32 {
+impl<'p> Pseudoheader for Ipv4Packet <'p> {
+    fn checksum(&self) -> u32 {
         let mut sum = 0u32;
 
         // Checksum pseudo-header
--- a/src/packet/ipv4.rs
+++ b/src/packet/ipv4.rs
@@ -39,22 +39,42 @@ pub struct Ipv4 {
     payload: Vec<u8>,
 }
 
+impl <'p> HasPseudoheader for Ipv4Packet <'p> {
+    fn pseudoheader_checksum(&self) -> u32 {
+        let mut sum = 0u32;
+
+        // Checksum pseudo-header
+        // IPv4 source
+        let source = self.get_source();
+        match source.octets() {
+            [a, b, c, d] => {
+                sum = sum + ((a as u32) << 8 | b as u32);
+                sum = sum + ((c as u32) << 8 | d as u32);
+            }
+        }
+
+        // IPv4 destination
+        let destination = self.get_destination();
+        match destination.octets() {
+            [a, b, c, d] => {
+                sum = sum + ((a as u32) << 8 | b as u32);
+                sum = sum + ((c as u32) << 8 | d as u32);
+            }
+        }
+
+        // IPv4 Next level protocol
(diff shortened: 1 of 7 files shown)


src/packet/ipv6.rs
```diff
@@ -22,7 +22,7 @@ pub struct Ipv6 {
     flow_label: u20be,
     payload_length: u16be,
     #[construct_with(u8)]
-    next_header: IpNextHeaderProtocol,
+    next_level_protocol: IpNextHeaderProtocol,
```
mrmonday: Why was this field renamed? It was intentionally named that way to match the specification.

src/packet/ipv6.rs
```diff
@@ -9,7 +9,7 @@
 //! IPv6 packet abstraction
 
 use packet::ip::IpNextHeaderProtocol;
-
+use packet::{Pseudoheader,PrimitiveValues};
```
mrmonday: Space after the comma, please.

src/packet/ipv4.rs
```diff
@@ -39,22 +39,37 @@ pub struct Ipv4 {
     payload: Vec<u8>,
 }
 
+impl<'p> Pseudoheader for Ipv4Packet <'p> {
```
mrmonday: No space between `Ipv4Packet` and `<'p>` please.

src/packet/mod.rs
```diff
@@ -10,6 +10,12 @@
 
 #![macro_use]
 
+/// Used to help compute checksum value of packet
+pub trait Pseudoheader {
```
mrmonday: Please make this CamelCase, so PseudoHeader, rather than Pseudoheader.

src/packet/udp.rs
```diff
@@ -101,6 +50,8 @@ fn udp_header_ipv4_test() {
     packet[20 + 8 + 3] = 't' as u8;
 
     {
+        use pnet::packet::ipv4::Ipv4Packet;
+        let ip_header = Ipv4Packet::new(&packet[..]).unwrap();
         let mut udp_header = MutableUdpPacket::new(&mut packet[20..]).unwrap();
```
mrmonday: So the error you're getting here:
```
src/packet/udp.rs:55:57: 55:63 error: cannot borrow `packet` as mutable because it is also borrowed as immutable
src/packet/udp.rs:55         let mut udp_header = MutableUdpPacket::new(&mut packet[20..]).unwrap();
                                                                             ^~~~~~
src/packet/udp.rs:54:42: 54:48 note: previous borrow of `packet` occurs here; the immutable borrow prevents subsequent moves or mutable borrows of `packet` until the borrow ends
src/packet/udp.rs:54         let ip_header = Ipv4Packet::new(&packet[..]).unwrap();
                                                              ^~~~~~
src/packet/udp.rs:70:6: 70:6 note: previous borrow ends here
src/packet/udp.rs:52     {
...
src/packet/udp.rs:70     }
                         ^
```
Is exactly as it says - you cannot have both mutable and immutable references to a packet at the same time. This is the reason the old `checksum` function took a source and destination address, rather than taking an `Ipv4Packet`.

There are various ways to work around this, I don't know which is best without some additional thought though.

mrmonday: I should note that the other errors you're getting are caused for the same reason. Since this is just a unit test, you could cheat and use http://doc.rust-lang.org/std/primitive.slice.html#method.split_mut - ultimately I think there needs to be a better solution though.

src/packet/ipv6.rs
david415: It was renamed so that ipv4 and ipv6 will have the same method name for getting the IpNextHeaderProtocol... However it doesn't have to be done this way. Instead we could abstract this away in the implementation of the Pseudo Header trait.

src/packet/ipv6.rs
david415: ok

src/packet/ipv4.rs
david415: noted

src/packet/mod.rs
david415: ok

david415:
no

PR #105, opened 2015-08-22
91.add modular checksum.2

david415:
This changeset has some more corrections from our last code review... however i still need to fix the unit tests; I'll do so in this branch. I'm looking at the rust docs about spliting slices...  http://doc.rust-lang.org/std/primitive.slice.html#method.split_mut

### DIFF
--- a/src/util.rs
+++ b/src/util.rs
@@ -11,6 +11,7 @@
 use bindings::libc;
 
 use packet::PrimitiveValues;
+use packet::PseudoHeader;
 
 use std::fmt;
 use std::str::FromStr;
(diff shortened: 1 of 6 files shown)


david415:
oh dear it looks like my bitwise arithmetic is wrong because the checksum tests fail. i'll come back to this later... frustrating.

david415:
well now... it doesn't matter if it won't merge because the bitwise arithmetic being wrong is actually the problem here. i guarantee that resolving the merge conflict will be way easier.

little-dude:
@david415 do you have any plan to continue working on this? If not, may I continue?

src/util.rs
```diff
@@ -355,3 +356,33 @@ fn get_network_interfaces_impl() -> Vec<NetworkInterface> {
...
+    let mut sum = initial;
+
+    let mut i = 0;
+    while i < length {
+        let word = (packet[i] as u32) << 8 | packet[i + 1] as u32;
+        sum = sum + word;
+        i = i + 2;
+    }
+    if packet.len()%2 == 1 {
+        sum = sum + (packet[length] as u32) << 8
```
little-dude: @david415 : I _think_ this is equivalent to `sum = (sum + (packet[length] as u32)) << 8` (at least that's what `clippy` suggested to me), whereas we want to do `sum = sum + ((packet[length] as u32) << 8)`:

```
src/util.rs:399:15: 399:49 warning: operator precedence can trip the unwary. Consider parenthesizing your expression:`(sum + (packet[length] as u32)) << 8`, #[warn(precedence)] on by default
src/util.rs:399         sum = sum + (packet[length] as u32) << 8
```

Also, `;` is missing, so the result is returned before the carries are added.

So changing this line `sum = sum + ((packet[length] as u32) << 8);` makes the following test pass:

```rust
#[test]
fn checksum_odd() {
    let data = vec![
        0x03, 0x01, 0x00, 0x00, 0x12, 0x34, 0x56, 0x78,
        0x01, 0x02, 0x03, 0x04, 0x05, 0x06, 0x07, 0x08,
        0x55];
    //   03 01
    // + 00 00
    // + 12 34
    // + 56 78
    // + 01 02
    // + 03 04
    // + 05 06
    // + 07 08
    // + 55 00
    //  +2 +2
    // --------
    //   d0 c1 = 1101 0000 1100 0001
    //
    // => checksum is 0010 1111 0011 1110
    //                   2    f    3    e
    assert_eq!(rfc1071_checksum(&data[..], 0), 0x2f3e);
}
```

david415:
wow nice catch! yes please continue working on this.

little-dude:
Cool, I'll keep you updated :)

mrmonday:
Hi,

I'm closing this pull request since it doesn't build and there has been no activity since last year. Please feel free to reopen it once it's building again!

PR #106, opened 2015-08-23
cleanup & refactoring

darxriggs:
NOTE: I can only test on Linux.


src/datalink/mod.rs
```diff
@@ -76,7 +76,7 @@ pub fn datalink_channel(network_interface: &NetworkInterface,
 /// Structure for sending packets at the data link layer. Should be constructed using
 /// datalink_channel().
 pub struct DataLinkSender {
-    dlsi: backend::DataLinkSenderImpl
+    sender: backend::DataLinkSenderImpl
```
mrmonday: I think this would be better named `backend`, and the same with the `DataLinkReceiver` struct.

src/datalink/mod.rs
```diff
@@ -119,21 +119,21 @@ impl DataLinkReceiver {
...
-            imp: self.dlri.iter()
+            channel_iterator: self.receiver.iter()
         }
     }
 }
 
 /// An iterator over data link layer packets
 pub struct DataLinkChannelIterator<'a> {
-    imp: backend::DataLinkChannelIteratorImpl<'a>,
+    channel_iterator: backend::DataLinkChannelIteratorImpl<'a>,
```
mrmonday: Again, I think this would be better named `backend` - it's a little confusing to have a `channel_iterator` inside a `channel_iterator`. Perhaps it would be worth looking at what other projects call this member, I know other projects use similar techniques.

mrmonday:
Other than the two small comments I've made, this all looks good, thank you! I'll merge it once you've addressed them.

The tests are automatically run on Linux and OS X - if they pass, that's good enough for me. I don't think Windows builds at all at the moment, if it does these changes won't affect that. No FreeBSD code is affected by this change, so that shouldn't be an issue.

src/datalink/mod.rs
darxriggs: Another suggestion is `delegate`. But I will go with your suggestion.

src/datalink/mod.rs
```diff
@@ -68,7 +68,7 @@ pub fn datalink_channel(network_interface: &NetworkInterface,
     -> io::Result<(DataLinkSender, DataLinkReceiver)> {
     match backend::datalink_channel(network_interface, write_buffer_size, read_buffer_size,
                                              channel_type) {
-        Ok((tx, rx)) => Ok((DataLinkSender { dlsi: tx }, DataLinkReceiver { dlri: rx })),
+        Ok((sender, receiver)) => Ok((DataLinkSender { backend: sender}, DataLinkReceiver { backend: receiver })),
```
mrmonday: `DataLinkSender { backend: sender}` => `DataLinkSender { backend: sender }` (notice the extra space).

src/util.rs
```diff
@@ -147,14 +143,14 @@ fn sockaddr_to_network_addr(sa: *const libc::sockaddr) -> (Option<MacAddr>, Opti
             let mac = MacAddr((*sll).sll_addr[0], (*sll).sll_addr[1],
                               (*sll).sll_addr[2], (*sll).sll_addr[3],
                               (*sll).sll_addr[4], (*sll).sll_addr[5]);
-            return (Some(mac), None);
+            (Some(mac), None)
```
mrmonday: When using implicit return, could you put an empty line before it, please? eg.

```
some_code;

return_val
```

Rather than:

```
some_code;
return_val
```

src/util.rs
```diff
@@ -147,14 +143,14 @@ fn sockaddr_to_network_addr(sa: *const libc::sockaddr) -> (Option<MacAddr>, Opti
...
         } else {
             let addr = internal::sockaddr_to_addr(mem::transmute(sa),
                                         mem::size_of::<libc::sockaddr_storage>());
-            return match addr {
+            match addr {
```
mrmonday: Same again here, please add an empty line before it.

mrmonday:
Again, I'm happy to merge this once you've fixed the issues I've mentioned. There's become quite a large number of commits, it might be nice if you could squash them, once you're done making changes.

src/datalink/mod.rs
darxriggs: Looks like I have found a companion. :-)
I am also that strict when it comes to code style.

mrmonday: I usually try to be fairly lenient with most style things, little things like this get me though!

I'm being stricter with this pull request too, with it being cleanup/refactoring.

mrmonday:
This all looks great, thanks for the cleanup!

@homu r+

mrmonday:
@homu retry

PR #107, opened 2015-08-25
FIXED Needs to with with iproute2 too

construidor:


mrmonday:
This looks great, thank you for your contribution!

@homu r+

PR #111, opened 2015-09-19
Remove the lint pass.

mrmonday:
Since it was unused anyway, and cause build failure on the latest Rust
nightlies.


mrmonday:
@homu r+

Issue #112, opened 2015-09-20
length_fn compile error 


charloco:
I have the following test struct

#[packet]
pub struct MsgTest{

    source: u16be,

    destination: u16be,

    #[length_fn="msg_options_length"]

    options:Vec<u8>,

    #[payload]
    payload: Vec<u8>
}
this works well.But when I add "    t:u8," after options:Vec<u8> just as following codes:
pub struct MsgTest{

    source: u16be,

    destination: u16be,

    #[length_fn="msg_options_length"]

    options:  Vec <u8>,

    t:u8, //this cause compile error

    #[payload]
    payload: Vec<u8>
}

the error message is 

<quote expansion>:141:47: 141:58 error: use of moved value: `self.packet` [E0382]
<quote expansion>:141         self_.packet[4 + msg_options_length(&self.to_immutable()) + 0] = (val) as u8;

mrmonday:
It's hard to tell you what the issue is you're encountering without seeing the code that's causing it - by the looks of things it's certainly not an issue with libpnet.

There's more information about the error you're getting, and how to fix it, here: http://doc.rust-lang.org/nightly/error-index.html#E0382

If you post a more complete version of your code I'd be happy to help you fix it.

charloco:
in fact,i am trying to write a mqtt server with libpnet.So i do some test first.

I create a libpnet/src/packet/mqtt.rs

and give the test struct 

```
#[packet]
pub struct Mqtt {
    source: u16be,
    destination: u16be,
    //length: u16be,
   // checksum: u16be,
    #[length_fn="mqtt_options_length"]
    options:Vec<u8>,
    t:u8,
    #[payload]
    payload: Vec<u8>
}
```
I guess that between "length_fn" and "payload" any field should not insert,

or can say that length_fn cann't be placed any place.

the full error msg

```
<quote expansion>:141:47: 141:51 error: use of moved value: `*self` [E0382]
<quote expansion>:141         self_.packet[4 + mqtt_options_length(&self.to_immutable()) + 0] = (val) as u8;
                                                                    ^~~~
/home/yuchang/work/git/rust/libpnet/src/packet/mqtt.rs:31:1: 31:10 note: in expansion of #[_packet_generator]
/home/yuchang/work/git/rust/libpnet/src/packet/mqtt.rs:31:1: 31:10 note: expansion site
<quote expansion>:141:47: 141:51 help: run `rustc --explain E0382` to see a detailed explanation
[...]

PR #113, opened 2015-09-21
Fix issue #112.

mrmonday:
The primary issue was that accessors and mutators could attempt to borrow
mutably and immutably at the same time if there was a length method specified.
This hadn't been caught previously due to no code having fields after a
variable length field.


mrmonday:
@homu r+

PR #114, opened 2015-10-08
Use traits to define DataLink*

rbran:
Using **Polymorphism** instead of **Encapsulation** is the key to use multiples backeds at the same time. Lets say that you need to open an datalink to a interface using bpf and an datalink on other interface using netmap, with this `trait` you can :)
Note1: I couldn't compile using netmap, so I didn't test using multiples backeds.
Note2: try to find a better solution to `default_datalink_channel` implementation

### DIFF
--- a/src/datalink/bpf.rs
+++ b/src/datalink/bpf.rs
@@ -19,17 +19,18 @@ use std::sync::Arc;
 use bindings::bpf;
 use packet::Packet;
 use packet::ethernet::{EthernetPacket, MutableEthernetPacket};
-use datalink::DataLinkChannelType;
+use datalink::{DataLinkChannelType, DataLinkSender, DataLinkReceiver, DataLinkChannelIterator};
 use datalink::DataLinkChannelType::{Layer2, Layer3};
 use internal;
 use util::NetworkInterface;
 
 // NOTE buffer must be word aligned.
+#[inline]
 pub fn datalink_channel(network_interface: &NetworkInterface,
                         write_buffer_size: usize,
                         read_buffer_size: usize,
                         channel_type: DataLinkChannelType)
-    -> io::Result<(DataLinkSenderImpl, DataLinkReceiverImpl)> {
+    -> io::Result<(Box<DataLinkSender>, Box<DataLinkReceiver>)> {
     #[cfg(target_os = "freebsd")]
     fn get_fd() -> libc::c_int {
         unsafe {
@@ -139,16 +140,16 @@ pub fn datalink_channel(network_interface: &NetworkInterface,
     }
 
     let fd = Arc::new(internal::FileDesc { fd: fd });
-    let sender = DataLinkSenderImpl {
+    let sender = Box::new(DataLinkSenderImpl {
         fd: fd.clone(),
         write_buffer: repeat(0u8).take(write_buffer_size).collect(),
         header_size: header_size,
-    };
-    let receiver = DataLinkReceiverImpl {
+    });
+    let receiver = Box::new(DataLinkReceiverImpl {
         fd: fd,
         read_buffer: repeat(0u8).take(read_buffer_size).collect(),
         header_size: header_size,
-    };
+    });
 
     Ok((sender, receiver))
 }
@@ -159,10 +160,10 @@ pub struct DataLinkSenderImpl {
     header_size: usize,
 }
 
-impl DataLinkSenderImpl {
-    pub fn build_and_send<F>(&mut self, num_packets: usize, packet_size: usize,
-                          func: &mut F) -> Option<io::Result<()>>
-        where F : FnMut(MutableEthernetPacket)
+impl DataLinkSender for DataLinkSenderImpl {
+    #[inline]
+    fn build_and_send(&mut self, num_packets: usize, packet_size: usize,
+                          func: &mut FnMut(MutableEthernetPacket)) -> Option<io::Result<()>>
     {
         let len = num_packets * (packet_size + self.header_size);
         if len >= self.write_buffer.len() {
@@ -193,7 +194,8 @@ impl DataLinkSenderImpl {
         }
     }
 
-    pub fn send_to(&mut self, packet: &EthernetPacket, _dst: Option<NetworkInterface>)
+    #[inline]
+    fn send_to(&mut self, packet: &EthernetPacket, _dst: Option<NetworkInterface>)
         -> Option<io::Result<()>> {
         match unsafe { libc::write(self.fd.fd,
                                    packet.packet().as_ptr() as *const libc::c_void,
@@ -210,14 +212,14 @@ pub struct DataLinkReceiverImpl {
     header_size: usize,
 }
 
-impl DataLinkReceiverImpl {
-    pub fn iter<'a>(&'a mut self) -> DataLinkChannelIteratorImpl<'a> {
+impl DataLinkReceiver for DataLinkReceiverImpl {
+    fn iter<'a>(&'a mut self) -> Box<DataLinkChannelIterator + 'a> {
         let buflen = self.read_buffer.len();
-        DataLinkChannelIteratorImpl {
+        Box::new(DataLinkChannelIteratorImpl {
             pc: self,
             // Enough room for minimally sized packets without reallocating
             packets: VecDeque::with_capacity(buflen / 64)
-        }
+        })
     }
 }
 
@@ -226,8 +228,8 @@ pub struct DataLinkChannelIteratorImpl<'a> {
     packets: VecDeque<(usize, usize)>,
 }
 
-impl<'a> DataLinkChannelIteratorImpl<'a> {
-    pub fn next<'c>(&'c mut self) -> io::Result<EthernetPacket<'c>> {
+impl<'a> DataLinkChannelIterator<'a> for DataLinkChannelIteratorImpl<'a> {
+    fn next<'c>(&'c mut self) -> io::Result<EthernetPacket<'c>> {
         if self.packets.is_empty() {
             let buflen = match unsafe {
                 libc::read(self.pc.fd.fd,
--- a/src/datalink/mod.rs
+++ b/src/datalink/mod.rs
@@ -9,7 +9,6 @@
 //! Support for sending and receiving data link layer packets
 
 use std::io;
-use std::iter::Iterator;
 use std::option::Option;
 
 use packet::ethernet::{EtherType, EthernetPacket, MutableEthernetPacket};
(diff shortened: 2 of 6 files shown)


mrmonday:
I think you might be my new favourite person. I've been meaning to make changes like this (and others) for a long time now! I'll take a look through now.

src/datalink/bpf.rs
```diff
@@ -139,30 +139,29 @@ pub fn datalink_channel(network_interface: &NetworkInterface,
...
     header_size: usize,
 }
 
-impl DataLinkSenderImpl {
-    pub fn build_and_send<F>(&mut self, num_packets: usize, packet_size: usize,
-                          func: &mut F) -> Option<io::Result<()>>
-        where F : FnMut(MutableEthernetPacket)
+impl DataLinkSender for DataLinkSenderBpf {
+    fn build_and_send(&mut self, num_packets: usize, packet_size: usize,
+                          func: &mut FnMut(MutableEthernetPacket)) -> Option<io::Result<()>>
```
mrmonday: This change worries me a lot - `build_and_send` now requires a dynamic dispatch, rather than static, which has the potential to kill performance - this is not so much an issue for the BPF backend, but for the likes of netmap, and potential DPDK or PF_RING ZC backends where it's possible to work with 10GbE links and every cycle counts this could be a killer.

The reason it was done the old way was to avoid this issue. Perhaps there is a happy middle ground we could find? (Or, alternatively, if you can give before/after assembly proving that the dynamic dispatch doesn't happen in practice, that would be fine too).

src/datalink/bpf.rs
```diff
@@ -19,17 +19,17 @@ use std::sync::Arc;
...
 use packet::ethernet::{EthernetPacket, MutableEthernetPacket};
-use datalink::DataLinkChannelType;
+use datalink::{DataLinkChannelType,DataLinkSender,DataLinkReceiver,DataLinkChannelIterator};
 use datalink::DataLinkChannelType::{Layer2, Layer3};
 use internal;
 use util::NetworkInterface;
 
 // NOTE buffer must be word aligned.
-pub fn datalink_channel(network_interface: &NetworkInterface,
+pub fn datalink_channel_bpf(network_interface: &NetworkInterface,
```
mrmonday: There's no need for the `_bpf` and `Bpf` suffixes to the various functions and structures - the module system can handle this.

src/datalink/bpf.rs
```diff
@@ -19,17 +19,17 @@ use std::sync::Arc;
 use bindings::bpf;
 use packet::Packet;
 use packet::ethernet::{EthernetPacket, MutableEthernetPacket};
-use datalink::DataLinkChannelType;
+use datalink::{DataLinkChannelType,DataLinkSender,DataLinkReceiver,DataLinkChannelIterator};
```
mrmonday: Spaces after the commas please.

Relatedly, I hope to integrate `rustfmt` into the build process in future so that style issues like this don't matter.

src/datalink/mod.rs
```diff
@@ -65,21 +53,54 @@ pub fn datalink_channel(network_interface: &NetworkInterface,
...
+                        write_buffer_size: usize, read_buffer_size: usize,
+                        channel_type: DataLinkChannelType)
+    -> io::Result<(Box<DataLinkSender>, Box<DataLinkReceiver>)> {
+    netmap::datalink_channel_netmap(network_interface, write_buffer_size,
+        read_buffer_size, channel_type)
+}
+
+/// Structure for sending packets at the data link layer. Should be constructed using
+/// datalink_channel().
+pub trait DataLinkSender: Send{
```
mrmonday: Spaces before the `:` and `{`, please.

mrmonday: Same for the other traits too, actually :)

src/datalink/mod.rs
```diff
@@ -100,38 +117,24 @@ impl DataLinkSender {
...
-impl DataLinkReceiver {
+pub trait DataLinkReceiver: Send {
     /// Returns an iterator over `EthernetPacket`s.
     ///
     /// This will likely be removed once other layer two types are supported.
     #[inline]
-    pub fn iter<'a>(&'a mut self) -> DataLinkChannelIterator<'a> {
-        DataLinkChannelIterator { backend: self.backend.iter() }
-    }
+    fn iter<'a>(&'a mut self) -> Box<DataLinkChannelIterator +'a>;
```
mrmonday: Space before the `'a`, please.

src/datalink/bpf.rs
rbran: I was thinking in make easy to differentiate names on import, like:

```
use pnet::datalink::netmap::{datalink_channel_netmap};
use pnet::datalink::linux::{datalink_channel_linux};

....

// Create a channel to receive on Interface 1
    let (_, mut rx1) = match datalink_channel_netmap(&interface1, 0, 4096, Layer2) {
        Ok((tx, rx)) => (tx, rx),
        Err(e) => panic!("packetdump: unable to create channel: {}", e)
    };

// Create a channel to receive on Interface 2
    let (_, mut rx2) = match datalink_channel_linux(&interface2, 0, 4096, Layer2) {
        Ok((tx, rx)) => (tx, rx),
        Err(e) => panic!("packetdump: unable to create channel: {}", e)
    };
```

mrmonday:
Other than the stylistic changes I've mentioned, this all looks good.

My big worry, as mentioned in another comment, is the overhead of dynamically dispatching when sending packets. Whilst this overhead is acceptable on 1GbE links, as you'd be targeting with WinPcap, BPF, and Linux backends, it's a killer when you start working with 10GbE and above with netmap (and potential future backends), where you have very few cycles for getting your processing done.

src/datalink/bpf.rs
mrmonday: The trick here is:

```rust
use pnet::datalink::netmap;

netmap::datalink_channel(...)
```

I would eventually also like to drop the `datalink_` prefix, but that needs greater design thought (I want to get rid of most of the parameters and just pass in an interface and some configuration, I haven't figured out the right way to do that though - may as well do both breaking changes at once).

rbran:
About the problems:
* Syntax: Solved, there something else? I'm not used to write in Rust.
* Suffixes: Solved (reverted).
* Dynamic dispatch: Unsolved :(

Unfortunately I don't know any solution for it right now, once that generics isn't a option on traits, but maybe make a benchmark before will better, we will know how much it really affect the performance.

mrmonday:
One option is something like:

```rust
pub fn build_and_send<F>(dls: &mut DataLinkSender,
                         num_packets: usize,
                         packet_size: usize,
                         func: &mut F)
 -> Option<io::Result<()>> { ... }
```

That is, having it as a free function - at that point you're more or less back where we started, but with a worse interface unfortunately. You could forward the trait through to the free function, but then the nicer interface has worse performance, which would be a shame.

I'm not sure what the right solution is here.

rbran:
I made a very simple benchmark to test dynamic vs static dispatch and added the possibility of function pointer, here the results:
https://gist.github.com/rbran/c16541d7c396dc874510

So @moonpolysoft I belive the function pointer be the best (even if you can't use closures), but I will leave for you to decide.

mrmonday:
I'm weary of these micro-benchmarks - they don't say anything for how this would work in a real world scenario. I think what I'd want to see is the performance of the minimum sized packet benchmarks with each of these methods - perhaps you could try that, using the loopback interface (you'll likely need to play with the numbers to make sure it runs long enough to be useful).

rbran:
@ mrmonday  after run `benches/rs_sender.rs` against `benches/rs_receiver.rs` on loop interface, using linux backed (unfortunately netmap didn't work), here the results:
```
* **Dynamic Dispatch**
Test1 Average=883485.05
Test2 Average=867778.16
```

```
* **Static Dispatch**
Test1 Average=872247.36
Test2 Average=883816.20
```

The data is very poor, inconclusive. I will try make netmap work, and bring better results.

By the way the free function its basically impractical, because it can't receive `DataLinkSender`, it has to by `DataLinkSenderImpl`, otherwise there is no way to access the internals of the struct (like self.write_buffer).

rbran:
@mrmonday I tried, but netmap backed don't work on linux, it send the packages, but send (always) a 2046 length packages with nothing but 0s (`build_and_send` and `send_to`). I'm tired and it's weekend, next week I do a more detailed debug :)

Here the code used to test (It's ugly but work for testing).
https://github.com/rbran/libpnet/tree/netmap-working-dynamic-dispatch
https://github.com/rbran/libpnet/tree/netmap-working-static-dispatch

mrmonday:
Sorry for my slow responses - I have very limited internet access at the moment.

I'm fairly sure netmap doesn't work on the loopback interface out of the box, you'll need to play around with VALE or netmap pipes I think - I haven't done this myself, so I'm not sure.

Are you running other things on your machine? Those numbers are incredibly low for the loopback interface. I'm unable to test to see what I'm getting right now, but ~880_000pps at 64 bytes per packet is only ~54MB/s - I'd expect far higher than that, even using the Linux backend.

rbran:
Netmap works on loopback interface, I tried with pkg-gen (part of netmap package) and get ~980Mbps (~1900Mpkts) with 64b packets.

About the Linux backed, I will investigate.

By the way, I'm out of internet too (using phone right now)

rbran:
I changed the distribution to kali linux (theoretically optimized), and write a C program with send package via raw socket, the speed was about ~60-70MB/s.

The real problem here is the pkg-gen (https://github.com/luigirizzo/netmap/blob/master/examples/pkt-gen.c) send packets to loopback and rs_sender.rs do not (at least not correctly).

mrmonday:
This is unusual - I don't know what's going on there. From a quick skim of the pkt-gen source, I can't see any difference in how we send packets - perhaps there is something missing though?

rbran:
On a quick debug I found out some strange behavior, if you put `build_and_send` (rs_sender) out of the loop it don't send nothing, but if you put it inside a loop, it will send garbage.

Let's say that I run `pkt-gen` (sends a very specific packet), when running `rs_sender` it will send the `pkt-gen` packets, the garbage on the ring (I believe)

By the way this pull request it's getting out of hand, first it was a change suggestion, then was need a benchmark, now we a trying to solve netmap related bug.

mrmonday:
Sounds like memory corruption - really need to audit the `unsafe` blocks in libpnet at some point! Would be great if there were an automated tool for it. I seem to recall fixing some issues with the offset functions in netmap_sys at some point, it doesn't look like I have a copy of those changes though (if I even made them).

Rather than delving further down the rabbit hole, you could look at a before/after dump of the assembly (or LLVM IR) for comparison - if LLVM is able to inline it either way then there's no need to run the benchmarks.

rbran:
I made a diff of the LLVM IR and Assembly, but I was not able to conclude anything, the code in "static-dispatch" is too much optimized for me to find even the function `build_and_send`!

See for yourself, if you're brave:
https://github.com/rbran/libpnet/blob/netmap-working-dynamic-dispatch/tmp/
https://github.com/rbran/libpnet/blob/netmap-working-static-dispatch/tmp/

mrmonday:
Renaming the main functions so they match, then using llvm-diff shows trivial differences in `fn main()` - that'll do me in terms of proof that they perform similarly! If it turns out to be a problem later on we can deal with it then.

I don't have time to review your code properly tonight (limited internet etc...) I'll take a look through tomorrow though and hopefully get this merged, if I can! Thanks for spending your time on this :)

src/datalink/bpf.rs
```diff
@@ -159,10 +159,9 @@ pub struct DataLinkSenderImpl {
     header_size: usize,
 }
 
-impl DataLinkSenderImpl {
-    pub fn build_and_send<F>(&mut self, num_packets: usize, packet_size: usize,
-                          func: &mut F) -> Option<io::Result<()>>
-        where F : FnMut(MutableEthernetPacket)
+impl DataLinkSender for DataLinkSenderImpl {
+    fn build_and_send(&mut self, num_packets: usize, packet_size: usize,
```
mrmonday: Could you mark all the `build_and_send` functions in the impls as `#[inline]` please? This should maximise the chance of the builder function getting inlined.

mrmonday:
This all looks great, thanks for your work! Two things before I merge:

 * Could you add the `#[inline]` attributes as I specified, please?
 * Could you squash your commits, however you feel is appropriate to do so (no need to have bug fix commits for bugs which were never released!)

With these two tweaks I'll get this merged in :)

mrmonday:
Ooh, one more thing actually, could you bump the version in Cargo.toml to 0.3.0? Thanks!

rbran:
Ok, squash to just one commit?

mrmonday:
That'll do - thank you!

@homu r+

PR #116, opened 2015-10-16
use libc::size_t instead of u64

aleksander:
This will fix compilation errors on 32bit non-windows platforms.


mrmonday:
Thank you for this! :)

@homu r+

PR #121, opened 2015-11-10
Fix Windows support.

mrmonday:


mrmonday:
Some tests are currently disabled in AppVeyor, they need some investigation into why they don't work there. Note that they do work locally.

@homu r+

PR #123, opened 2015-11-11
Workaround rust-lang-nursery/libc#58.

mrmonday:
This fixes libpnet to an older version of libc, until this is fixed.

See also: rust-lang-nursery/libc#58.


mrmonday:
@homu r+

PR #126, opened 2015-11-20
Update to libc 0.2.

mrmonday:


mrmonday:
So it would seem libc 0.2 removed pretty much all support for Windows, so I'm going to put this off, since it seems you can't just drop winapi in there as a replacement.

mrmonday:
See also: rust-lang-nursery/libc#45 - it would seem it's going to be a ton of effort to write anything cross platform going forwards if we keep relying on libc - we probably need to look into writing a wrapper crate of some kind.

PR #127, opened 2015-11-22
remove clippy needless_lifetimes attribute

little-dude:
https://github.com/Manishearth/rust-clippy/issues/417 is now solved, so
this attribute is not required anymore.

### DIFF
--- a/src/datalink/bpf.rs
+++ b/src/datalink/bpf.rs
@@ -237,8 +237,6 @@ pub struct DataLinkReceiverImpl {
 }
 
 impl DataLinkReceiver for DataLinkReceiverImpl {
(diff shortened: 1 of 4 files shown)


mrmonday:
@homu r+

Thank you, it's always great to remove a FIXME! :D

little-dude:
> Thank you, it's always great to remove a FIXME! :D

Yeah it's pretty minor but it feels good :smile:

mrmonday:
I've just noticed actually, it would seem that clippy has never been run for `--features netmap` (only noticed this because the pull request didn't modify `src/datalink/netmap.rs`). I should really add some continuous integration for netmap...

Issue #128, opened 2015-11-22
u16 for header fields in a packet should trigger an error


little-dude:
From https://github.com/libpnet/libpnet/pull/89#issuecomment-158792770
Compiler should not accept `u16` attribute fields in a packet. Only  `u16be` or `u16le` are accepted. 

```rust
#[packet]
pub struct MyPacket {
   a_field: u16,  // should be refused
}
```

mrmonday:
I never can remember the format for commit messages to make this happen automatically... Fixed now though!

PR #129, opened 2015-11-22
Fix issue #128.

mrmonday:
Endianness must now be specified for fields in packets, if their size is
greater than eight.


mrmonday:
@homu r+

PR #136, opened 2015-12-05
Fix travis build.

mrmonday:
Travis recently removed support for IPv6 networking, so, consequently, the IPv6
tests are failing.


mrmonday:
@homu r+

Issue #140, opened 2015-12-06
Simple packet composition


mrmonday:
Currently, `benches/rs_sender.rs` is 112 lines long, compared to just 22 of `benches/py_sender.py`. It is possible to massively reduce the amount of code needed to compose packets in Rust, and this should be done.

pvinci:
I can see that this has been outstanding for a long while, but what are you thoughts on how to approach this?
Something like this looks like idiomatic rust and friendly to anyone who has used scapy.

```rust
let ether = Ether! (dst = "11:22:33:44:55:66", type = 0x0800);  // type should not be set, but be populated by build! from IP! 
let ip = IP! (ttl = 4, proto = 17, src = "127.0.0.1", dst = "127.0.0.1");  // proto should not be set but be populated by UDP!
let udp = UDP! (sport = 1234, dport = 1234);

let buffer = build! (ether / ip / udp / b"payload");

let buffer = build!( Ether! (dst = "11:22:33:44:55:66") / IP! (ttl = 4, src = "127.0.0.1", dst = "127.0.0.1") / UDP! (sport = 1234, dport = 1234 / b"payload");
```

infosechoudini:
Did we implement the above? Should we flesh out a feature request?

PR #146, opened 2015-12-12
checksum

little-dude:
still work in progress...

### DIFF
--- a/src/packet/ipv4.rs
+++ b/src/packet/ipv4.rs
@@ -11,7 +11,7 @@
 use packet::ip::IpNextHeaderProtocol;
 use packet::{PseudoHeader, PrimitiveValues};
 use util::rfc1071_checksum;
-use pnet_macros::types::*;
+use pnet_macros_support::types::*;
 
 use std::net::Ipv4Addr;
 
--- a/src/packet/ipv4.rs
+++ b/src/packet/ipv4.rs
@@ -9,7 +9,7 @@
 //! IPv4 packet abstraction
 
 use packet::ip::IpNextHeaderProtocol;
-use packet::{PseudoHeader, PrimitiveValues};
+use packet::{Packet, PseudoHeader, PrimitiveValues};
 use util::rfc1071_checksum;
 use pnet_macros_support::types::*;
 
@@ -41,37 +41,54 @@ pub struct Ipv4 {
     payload: Vec<u8>,
 }
 
+/// Represents an IPv4 Packet
+///
+/// 0          8          16          24         32
+/// +---------------------------------------------+
+/// |                  source_ip                  |
+/// +----------------------+----------------------+
+/// |                destination_ip               |
+/// +-----------+----------+----------------------+
+/// |   zeros   | protocol | tcp/udp length       |
+/// +-----------+---------------------------------+
+#[packet]
+pub struct Ipv4PseudoHeader {
+    #[construct_with(u8, u8, u8, u8)]
+    source: Ipv4Addr,
+    #[construct_with(u8, u8, u8, u8)]
+    destination: Ipv4Addr,
+    zeros: u8,
+    #[construct_with(u8)]
+    next_level_protocol: IpNextHeaderProtocol,
+    inner_packet_length: u16be,
+    // just to avoid the compiler to complain
+    #[payload]
+    payload: Vec<u8>,
+}
 
 impl<'p> PseudoHeader for Ipv4Packet<'p> {
-    fn checksum(&self) -> u32 {
-        let mut sum = 0u32;
 
-        // Checksum pseudo-header
-        // IPv4 source
-        let source = self.get_source();
-        match source.octets() {
-            [a, b, c, d] => {
-                let src = vec![a, b, c, d];
-                sum = (sum + (src[0] as u32) + (src[2] as u32)) << 8;
-                sum = sum + src[1] as u32 + src[3] as u32;
+    /// Return a PseudoHeader packet out of the packet header.
+    ///
+    /// The `inner_packet_length` is optional, since in case of UDP, it is taken from the length
+    /// field of the UDP header, whereas for TCP, it is computed from the header length.
+    fn get_pseudo_header(&self, inner_packet_length: Option<u32>) -> Vec<u8> {
+        let mut pseudo_header_buf: Vec<u8> = vec![0, 12];
+        {
+            let mut pseudo_header = MutableIpv4PseudoHeaderPacket::new(&mut pseudo_header_buf[..]).unwrap();
+            pseudo_header.set_source(self.get_source());
+            pseudo_header.set_destination(self.get_destination());
+            pseudo_header.set_next_level_protocol(self.get_next_level_protocol());
+            if let Some(inner_packet_length) = inner_packet_length {
+                pseudo_header.set_inner_packet_length(inner_packet_length as u16);
+            } else {
+                pseudo_header.set_inner_packet_length(
+                    // FIXME: is u16 legit for payload length?
+                    self.payload().len() as u16 - (4 as u16 * self.get_header_length() as u16)
+                );
             }
         }
-
-        // IPv4 destination
-        let destination = self.get_destination();
-        match destination.octets() {
-            [a, b, c, d] => {
-                let dst = vec![a, b, c, d];
-                sum = (sum + (dst[0] as u32) + (dst[2] as u32)) << 8;
-                sum = sum + dst[1] as u32 + dst[3] as u32;
-            }
-        }
-
-        // IPv4 Next level protocol
-        let next_level_protocol = self.get_next_level_protocol();
-        let (next_proto,) = next_level_protocol.to_primitive_values();
-        sum = sum + (next_proto as u32);
-        sum
+        pseudo_header_buf
     }
 }
 
@@ -82,8 +99,7 @@ fn ipv4_options_length(ipv4: &Ipv4Packet) -> usize {
 /// Calculates the checksum of an IPv4 packet
 pub fn checksum(packet: &Ipv4Packet) -> u16be {
     use packet::Packet;
-
-    rfc1071_checksum(packet.packet(), 0)
+    rfc1071_checksum(packet.packet(), None)
 }
 
 #[test]
--- a/src/packet/ipv4.rs
+++ b/src/packet/ipv4.rs
@@ -51,8 +51,9 @@ impl<'p> PseudoHeader for Ipv4Packet<'p> {
         let source = self.get_source();
         match source.octets() {
             [a, b, c, d] => {
-                sum = sum + ((a as u32) << 8 | b as u32);
-                sum = sum + ((c as u32) << 8 | d as u32);
+                let src = vec![a, b, c, d];
+                sum = sum + (src[0] as u32) + (src[2] as u32) << 8;
+                sum = sum + src[1] as u32 + src[3] as u32;
             }
         }
 
@@ -60,15 +61,16 @@ impl<'p> PseudoHeader for Ipv4Packet<'p> {
         let destination = self.get_destination();
         match destination.octets() {
             [a, b, c, d] => {
-                sum = sum + ((a as u32) << 8 | b as u32);
-                sum = sum + ((c as u32) << 8 | d as u32);
+                let dst = vec![a, b, c, d];
+                sum = sum + (dst[0] as u32) + (dst[2] as u32) << 8;
+                sum = sum + dst[1] as u32 + dst[3] as u32;
             }
         }
 
         // IPv4 Next level protocol
         let next_level_protocol = self.get_next_level_protocol();
         let (next_proto,) = next_level_protocol.to_primitive_values();
-        sum = sum + next_proto as u32;
+        sum = sum + (next_proto as u32);
         return sum;
     }
 }
--- a/src/packet/ipv4.rs
+++ b/src/packet/ipv4.rs
@@ -9,10 +9,12 @@
 //! IPv4 packet abstraction
 
 use packet::ip::IpNextHeaderProtocol;
-
-use pnet_macros_support::types::*;
+use packet::{PseudoHeader, PrimitiveValues};
+use util::rfc1071_checksum;
+use pnet_macros::types::*;
 
 use std::net::Ipv4Addr;
+
 
 /// Represents an IPv4 Packet
 #[packet]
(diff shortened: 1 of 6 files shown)


src/packet/ipv4.rs
```diff
@@ -39,29 +41,67 @@ pub struct Ipv4 {
...
-    }
-    while sum >> 16 != 0 {
-        sum = (sum >> 16) + (sum & 0xFFFF);
+impl<'p> PseudoHeader for Ipv4Packet<'p> {
+
+    /// Return a PseudoHeader packet out of the packet header.
+    ///
+    /// The `inner_packet_length` is optional, since in case of UDP, it is taken from the length
+    /// field of the UDP header, whereas for TCP, it is computed from the header length.
+    fn get_pseudo_header(&self, inner_packet_length: Option<u32>) -> Vec<u8> {
```
little-dude: I'd like to return a `Ipv4PseudoHeaderPacker` here, but rust won't compile, since the underlying vec that hold the data has a lifetime limited to this function.

As a result, end up writing this kind of thing: https://github.com/libpnet/libpnet/pull/146/files#diff-0739d94a24c78d29f3c7feb2669e2b26R29

```rust
let pseudo_header = ip_packet.get_pseudo_header(Some(udp_packet.get_length() as u32));
rfc1071_checksum(udp_packet.packet(), Some(&pseudo_header[..]))
// I'd like to be able to write this instead
// rfc1071_checksum(udp_packet.packet(), Some(pseudo_header.packet()))

File: src/util.rs [hunk 2]
```diff
@@ -419,3 +419,165 @@ fn get_network_interfaces_impl() -> Vec<NetworkInterface> {
 
     vec
 }
+
```
little-dude: Added a bunch of tests for the checksum function.

src/packet/ipv6.rs
```diff
@@ -32,6 +32,86 @@ pub struct Ipv6 {
...
+    type Item = &'h Ipv6Packet<'h>;
+    fn next(&mut self) -> Option<&'h Ipv6Packet<'h>> {
+        match self.current_header.get_next_header() {
+            IpNextHeaderProtocols::Hopopt |
+            IpNextHeaderProtocols::Ipv6 |
+            IpNextHeaderProtocols::Ipv6Route | 
+            IpNextHeaderProtocols::Ipv6Frag |
+            IpNextHeaderProtocols::Ipv6Icmp |
+            IpNextHeaderProtocols::Ipv6Opts => {
+                self.current_header = Ipv6Packet::new(self.current_header.payload()).unwrap(); // is it ok to unwrap here?
```
little-dude: This does not compile but I don't understand why:

```
/home/travis/build/libpnet/libpnet/src/packet/ipv6.rs:81:75: 81:84 error: no method named `payload` found for type `packet::ipv6::Ipv6Packet<'h>` in the current scope
```

src/packet/ipv6.rs
```diff
@@ -32,6 +32,86 @@ pub struct Ipv6 {
...
+    fn get_pseudo_header(&self, inner_packet_length: Option<u32>) -> Vec<u8> {
+        let mut pseudo_header_buf: Vec<u8> = vec![0, 40];
+        {
+            let mut pseudo_header = MutableIpv6PseudoHeaderPacket::new(&mut pseudo_header_buf[..]).unwrap();
+            pseudo_header.set_source(self.get_source());
+            // FIXME: not sure which header I should take for this one...
+            pseudo_header.set_destination(self.get_destination());
+            // Get the next_level_protocol and inner_packet_length from the last ipv6 header
+            loop {
+                if let Some(header) = self.next() {
```
little-dude:
```
/home/travis/build/libpnet/libpnet/src/packet/ipv6.rs:100:44: 100:50 error: no method named `next` found for type `&packet::ipv6::Ipv6Packet<'p>` in the current scope
```
I don't understand this one either...

little-dude:
@mrmonday apart from the compilation error, I'm not really satisfied by what I've done here... Any suggestion to improve?

src/packet/ipv4.rs
mrmonday: You won't be able to return an `Ipv4PseudoHeader` here, at least not with the current method signature.

The issue is that you're allocating a new vector, which the `get_pseudo_header()` function then owns. Either it needs to return it, so that the caller then owns it, or it needs to be be dropped at the end of the function.

The way around this (and what I'd like you to do - the heap allocation is unfortunate and unnecessary) is something like (untested, but it gives you the general idea):

```
fn populate(&self, buffer: &mut [u8], inner_packet_length: Option<usize>) -> Ipv4PseudoHeader; 
```

Here, the caller allocates the buffer, which the function then populates. Since the caller owns the buffer, its lifetime is managed by the caller, and the return value can outlive the function.

src/packet/ipv4.rs
```diff
@@ -39,29 +41,67 @@ pub struct Ipv4 {
...
+        {
+            let mut pseudo_header = MutableIpv4PseudoHeaderPacket::new(&mut pseudo_header_buf[..]).unwrap();
+            pseudo_header.set_source(self.get_source());
+            pseudo_header.set_destination(self.get_destination());
+            pseudo_header.set_next_level_protocol(self.get_next_level_protocol());
+            if let Some(inner_packet_length) = inner_packet_length {
+                pseudo_header.set_inner_packet_length(inner_packet_length as u16);
+            } else {
+                pseudo_header.set_inner_packet_length(
+                    // FIXME: is u16 legit for payload length?
```
mrmonday: Lengths for buffers should always be `usize`, unless the spec says otherwise (even then, usize is more convenient, unless you implement Index to avoid casting everywhere).

src/packet/ipv6.rs
mrmonday: You probably need to `use packet::Packet`.

Calling `.unwrap()` is safe, providing `self.current_header.payload().len() >= Ipv6Packet::minimum_packet_size()`.

Relatedly, though not particularly importantly, it's a shame you have to write out this code, it seems like something that pnet_macros should be able to do for you. There's currently functionality for handling optional fields (ie. IPv4 options), but obviously each potential next header field for IPv6 has a different packet type. It might be worth investigating at some point to see if it can be automated or not.

src/packet/ipv6.rs
mrmonday: `self` is an `&Ipv6Packet`, which has no method named next... I assume you're looking for the `next()` method in `Ipv6HeaderIter`?

src/packet/ipv4.rs
```diff
@@ -39,29 +41,67 @@ pub struct Ipv4 {
     payload: Vec<u8>,
 }
 
-/// Calculates the checksum of an IPv4 packet
-pub fn checksum(packet: &Ipv4Packet) -> u16be {
-    use packet::Packet;
+/// Represents an IPv4 Packet
```
mrmonday: Pseudo header, rather than packet, no?

src/packet/ipv4.rs
```diff
@@ -39,29 +41,67 @@ pub struct Ipv4 {
...
+/// +-----------+----------+----------------------+
+/// |   zeros   | protocol | tcp/udp length       |
+/// +-----------+---------------------------------+
+#[packet]
+pub struct Ipv4PseudoHeader {
+    #[construct_with(u8, u8, u8, u8)]
+    source: Ipv4Addr,
+    #[construct_with(u8, u8, u8, u8)]
+    destination: Ipv4Addr,
+    zeros: u8,
```
mrmonday: Are the zeros really necessary? I assume this is just for padding to get the protocol to be big endian?

src/packet/ipv4.rs
```diff
@@ -39,29 +41,67 @@ pub struct Ipv4 {
...
+    #[construct_with(u8)]
+    next_level_protocol: IpNextHeaderProtocol,
+    inner_packet_length: u16be,
+    // just to avoid the compiler to complain
```
mrmonday: No need for this comment... Though perhaps a bug report?

The idea of having the payload field was to support chaining different packet types together, it definitely feels out of place for certain packets though. I guess it will be investigated properly alongside issue #140.

src/packet/ipv4.rs
```diff
@@ -39,29 +41,67 @@ pub struct Ipv4 {
...
+            if let Some(inner_packet_length) = inner_packet_length {
+                pseudo_header.set_inner_packet_length(inner_packet_length as u16);
+            } else {
+                pseudo_header.set_inner_packet_length(
+                    // FIXME: is u16 legit for payload length?
+                    self.payload().len() as u16 - (4 as u16 * self.get_header_length() as u16)
+                );
+            }
+        }
+        pseudo_header_buf
```
mrmonday: Nitpick: I'd like a blank line before anything being returned without a return statement, please.

src/packet/ipv6.rs
```diff
@@ -32,6 +32,86 @@ pub struct Ipv6 {
     payload: Vec<u8>,
 }
 
+/// Represents an IPv6 Packet pseudo header
+#[packet] pub struct Ipv6PseudoHeader {
```
mrmonday: Newline between attributes and items please:

```rust
#[packet]
pub struct Ipv6PseudoHeader {
```

src/packet/ipv6.rs
```diff
@@ -32,6 +32,86 @@ pub struct Ipv6 {
...
+    #[construct_with(u16, u16, u16, u16, u16, u16, u16, u16)]
+    source: Ipv6Addr,
+    // RFC 2460:
+    // If the IPv6 packet contains a Routing header, the Destination Address used in the
+    // pseudo-header is that of the final destination. At the originating node, that address will
+    // be in the last element of the Routing header; at the recipient(s), that address will be in
+    // the Destination Address field of the IPv6 header.
```
mrmonday: These comments should probably be turned into doc comments - the compiler will likely complain once you've fixed the other compile errors.

src/packet/ipv6.rs
```diff
@@ -32,6 +32,86 @@ pub struct Ipv6 {
...
+    // RFC 2460: 
+    // The Upper-Layer Packet Length in the pseudo-header is the length of the
+    // upper-layer header and data (e.g., TCP header plus TCP data). Some upper-layer protocols
+    // carry their own length information (e.g., the Length field in the UDP header); for such
+    // protocols, that is the length used in the pseudo- header. Other protocols (such as TCP) do
+    // not carry their own length information, in which case the length used in the pseudo-header
+    // is the Payload Length from the IPv6 header, minus the length of any extension headers
+    // present between the IPv6 header and the upper-layer header.
+    inner_packet_length: u32be,
+    zeros: u24be,
```
mrmonday: If these zeros fields are sticking around, perhaps they should be renamed to `padding`, and have a comment added specifying why they're needed.

src/packet/ipv6.rs
```diff
@@ -32,6 +32,86 @@ pub struct Ipv6 {
...
+    // RFC 2460:
+    // The Next Header value in the pseudo-header identifies the upper-layer protocol (e.g., 6 for
+    // TCP, or 17 for UDP). It will differ from the Next Header value in the IPv6 header if there
+    // are extension headers between the IPv6 header and the upper-layer header.
+    #[construct_with(u8)]
+    next_level_protocol: IpNextHeaderProtocol,
```
mrmonday: Can this be renamed to `next_header`, to be consistent with the `Ipv6` packet, please?

src/packet/mod.rs
```diff
@@ -10,6 +10,22 @@
 
 #![macro_use]
 
+/// Used to help compute checksum value of packet
+pub trait PseudoHeader {
+    /// Converts IPv4 or IPv6 specific fields to a checksum value
+    // Should we add a "next_header" u16 argument for ipv6 packets with extension header?
```
mrmonday: I think it would be a shame to do anything protocol specific - ideally this abstraction should be protocol agnostic.

If it is 100% necessary, I wouldn't be opposed to IP specific functionality, given their ubiquity.

src/util.rs
```diff
@@ -419,3 +419,165 @@ fn get_network_interfaces_impl() -> Vec<NetworkInterface> {
...
+    let mut i = 0;
+    while i < length {
+        let word = (data[i] as u32) << 8 | data[i + 1] as u32;
+        sum = sum + word;
+        i = i + 2;
+    }
+    if data.len() % 2 == 1 {
+        sum = sum + ((data[length] as u32) << 8);
+    }
+    sum
```
mrmonday: Blank line before return please.

src/util.rs
mrmonday: I'm gonna trust that you've calculated these checksums properly, and that they are correct - were they made up by either you or @david415, or were from examples in the spec?

src/util.rs
```diff
@@ -419,3 +419,165 @@ fn get_network_interfaces_impl() -> Vec<NetworkInterface> {
...
+}
+
+
```
mrmonday: One too many new lines!

src/util.rs
mrmonday: I notice there are no tests with use `Some(_)` for the second parameter - could some be added? I know it's done as part of the other tests, I'd like there to be some checksum specific test cases though for completeness.

src/util.rs
```diff
@@ -419,3 +419,165 @@ fn get_network_interfaces_impl() -> Vec<NetworkInterface> {
...
+    //   7b c1 = 0111 1011 1100 0001
+    //
+    // => checksum is 1000 0100 0011 1110
+    //                   8    4    3    e
+    assert_eq!(rfc1071_checksum(&data[..], None), 0x843e);
+}
+
+#[test]
+fn checksum_even_with_carry() {
+    let data = vec![
```
mrmonday: I'd rather these `vec![]`s be static arrays instead, so `let data = [...];` - it doesn't really matter for test cases, but is more consistent with that other tests do.

The styling on these functions is a little weird. Perhaps, given the comments below, it might be better to write them as:
```rust
let data = [
    0x30, 0x10, //    30 10
    0x00, 0x00, // + 00 00
    ...
];
```
That said, I'm not sure repeating the numbers in comments is particularly necessary.

mrmonday:
Is there any particular reason you're not happy with this, @little-dude?

Other than the things I've pointed out:

 * Unnecessary allocation
 * Things which could possibly have some amount of auto-generation in future

It looks pretty good to me - I'd obviously give it a more thorough look over before merging, but there's nothing immediately obvious.

src/packet/mod.rs
david415: the pseudo header is specific to ipv4 and ipv6... so having this be a trait with two implementations makes sense.

src/packet/ipv4.rs
david415: that function signature won't work for our PseudoHeader trait because it must return a value which is suitable for ipv4 and ipv6... what should we do instead? perhaps an enum type which is either Ipv4PseudoHeader or Ipv6PseudoHeader?

mrmonday: An [associated type](http://doc.rust-lang.org/book/associated-types.html) could potentially work here instead. I haven't finished reading through the code yet though, an enum may be better.

little-dude: I tried associated types but got issues with lifetimes afterwards. I'll show the code when I am back to my laptop.

little-dude: @david415 : https://github.com/little-dude/libpnet/commit/a9bdc9298c076153fe84698846a1db7fb9914fae

here is my attempt to use associated types. Unfortunately it does not compile and I have no idea why...

sidenote: if me want the caller to allocate the buffer, I thought the IP packet could implement a `get_pseudo_header_buffer()` method, so that we don't always have to remember how big a buffer is. I tried with associated types too, but it does not work either: https://github.com/little-dude/libpnet/commit/2040487ec7b8cb0de9dde3018d3403539ea1f3b0

mrmonday:
Hi,

I'm closing this pull request since it doesn't build and there has been no activity since the start of the year. Please feel free to reopen it once it's building again!

Issue #137, opened 2015-12-06
Get libpnet building on stable Rust


mrmonday:
All the remaining features we rely on from the standard library have been either stabilised or deprecated for the 1.6 release cycle, with the exception of compiler plugins - we can use [syntex](https://github.com/serde-rs/syntex) to get around this issue.

mrmonday:
The first step towards this is done - as of 0.6.1, libpnet no longer requires any features other than plugins.

The next step is to switch over to syntex, so libpnet_macros can work with the beta release.

The final step will be waiting for the release of Rust 1.6.0, at which point libpnet should work on stable Rust.

PR #147, opened 2015-12-13
Introduce syntex to enable libpnet to work on beta Rust.

mrmonday:
This isn't ready for merging just yet - there are various cleanups required first. This is the penultimate step before #137 is complete.


mrmonday:
Woo hoo! This is the biggest hurdle for #137 complete. It should simply be a case of "enable CI on stable" when Rust 1.6 is released now.

Unfortunately this is going to break all other pull requests, I'll try and talk people through the changes they need to make once we see what's broken though.

@homu r+

PR #153, opened 2015-12-29
ipv4 lengths

alexsult:
Hi,                                                                                                                                                                                                                                
                                                                                                                                                                                                                                   
I have a small piece of code where I try to get the options of an IPv4 packet                                                                                                                                                      

    header = Ipv4Packet::new(ethernet.payload()).unwrap();
    let options = header.get_options();
    println!("Options {:?}",options);                                                                                                                                                                                              
                                                                                                                                                                                                                                   
And when it reads a packet,                                                                                                                                                                                                        
(built with scapy : 

    ip_packet=IP(src="1.1.1.1", dst="2.2.2.2", options=IPOption('\x83\x03\x10'))

)                                                                                                                      
                                                                                                                                                                                                                                   
I have the following message : 
                                                                                                                                                                                                    
    'slice index starts at 20 but ends at 1', ../src/libcore/slice.rs:518                                                                                                                                                              
                                                                                                                                                                                                                                   
So I modified the following files :                                                                                                                                                                                                
                                                                                                                                                                                                                                   
In src/packet/ipv4.rs:                                                                                                                                                                                                             
- I modified ipv4_options_length : the function did return the length in "words" instead of returning in bytes                                                                                                                     
- few ipv4 options are only 3 bytes long (EOL, NOP) and don't contain a length field. The easiest way (I found) to handle this, is to define that field (the length) as a Vector with a length deduced from the option number.     
- and I added a test for the ipv4 option packet                                                                                                                                                                                    
                                                                                                                                                                                                                                   
In pnet_macros/src/decorator.rs:                                                                                                                                                                                                   
- I modified the end of the slices to the "offset + length"  instead of "length".                                                                                                                                                  
                                                                                                                                                                                                                                   
I hope it's OK for you (and it helps), tell me if I can change anything,                                                                                                                                                           
                                                                                                                                                                                                                                   
Best regards,                                                                                                                                                                                                                      
                                                                                                                                                                                                                                   
Alexis

### DIFF
--- a/pnet_macros/src/decorator.rs
+++ b/pnet_macros/src/decorator.rs
@@ -474,7 +474,7 @@ fn handle_misc_field(cx: &mut GenContext,
     *mutators = format!("{mutators}
                     /// Set the value of the {name} field
                     #[inline]
-                    #[allow(trivial_numeric_casts)]
+                    #[allow(trivial_numeric_casts, used_underscore_binding)]
                     pub fn set_{name}(&mut self, val: {ty_str}) {{
                         use pnet::packet::PrimitiveValues;
                         let _self = self;
@@ -500,7 +500,7 @@ fn handle_misc_field(cx: &mut GenContext,
     *accessors = format!("{accessors}
                         /// Get the value of the {name} field
                         #[inline]
-                        #[allow(trivial_numeric_casts)]
+                        #[allow(trivial_numeric_casts, used_underscore_binding)]
                         pub fn get_{name}(&self) -> {ty_str} {{
                             {ctor}
                         }}
@@ -520,7 +520,7 @@ fn handle_vec_primitive(cx: &mut GenContext,
             *accessors = format!("{accessors}
                                     /// Get the value of the {name} field (copies contents)
                                     #[inline]
-                                    #[allow(trivial_numeric_casts)]
+                                    #[allow(trivial_numeric_casts, used_underscore_binding)]
                                     pub fn get_{name}(&self) -> Vec<{inner_ty_str}> {{
                                         let _self = self;
                                         let current_offset = {co};
@@ -548,7 +548,7 @@ fn handle_vec_primitive(cx: &mut GenContext,
         *mutators = format!("{mutators}
                                 /// Set the value of the {name} field (copies contents)
                                 #[inline]
-                                #[allow(trivial_numeric_casts)]
+                                #[allow(trivial_numeric_casts, used_underscore_binding)]
                                 pub fn set_{name}(&mut self, vals: Vec<{inner_ty_str}>) {{
                                     use std::ptr::copy_nonoverlapping;
                                     let mut _self = self;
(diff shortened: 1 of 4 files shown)


mrmonday:
Thank you for your pull request!

I don't have time to review it right now - I will try to do so this evening for you.

alexsult:
Thanks,

Alexis

src/packet/ipv4.rs
```diff
@@ -76,11 +79,28 @@ pub struct Ipv4Option {
...
     number: u5,
-    length: u8,
+    #[length_fn = "ipv4_options_length_length"]
+    length: Vec<u8>,
+    #[length_fn = "ipv4_options_payload_length"]
     #[payload]
     data: Vec<u8>,
 }
 
+fn ipv4_options_length_length(option: &Ipv4OptionPacket) -> usize {
```
mrmonday: Could you add a one-line comment for this function specifying what it does?

Also, perhaps a better name would be `ipv4_option_length`?

src/packet/ipv4.rs
```diff
@@ -76,11 +79,28 @@ pub struct Ipv4Option {
     copied: u1,
     class: u2,
     number: u5,
-    length: u8,
+    #[length_fn = "ipv4_options_length_length"]
```
mrmonday: Could you add a comment here specifying that this is an optional field, rather than a vector?

src/packet/ipv4.rs
```diff
@@ -76,11 +79,28 @@ pub struct Ipv4Option {
...
+    length: Vec<u8>,
+    #[length_fn = "ipv4_options_payload_length"]
     #[payload]
     data: Vec<u8>,
 }
 
+fn ipv4_options_length_length(option: &Ipv4OptionPacket) -> usize {
+    match option.get_number() {
+        0 => 0,
```
mrmonday: I would rather constants be used here, rather than integers - could you add something like `pnet::packet::ip::IpNextHeaderProtocols` and `pnet::packet::ethernet::EtherTypes`, and use constants defined in there instead?

The list doesn't need to be complete - I would like to avoid magic numbers though if possible.

src/packet/ipv4.rs
```diff
@@ -76,11 +79,28 @@ pub struct Ipv4Option {
...
+        1 => 0,
+        _ => 1,
+    }
+}
+
+fn ipv4_options_payload_length(ipv4_options: &Ipv4OptionPacket) -> usize {
```
mrmonday: Could you rename this to `ipv4_option_payload_length` (notice the lack of `s` after option).

pnet_macros/src/decorator.rs
```diff
@@ -596,9 +596,9 @@ fn handle_vector_field(cx: &mut GenContext,
                                 pub fn get_{name}_raw(&self) -> &[u8] {{
                                     let _self = self;
                                     let current_offset = {co};
-                                    let len = {packet_length};
+                                    let end = current_offset + {packet_length};
```
mrmonday: When making changes to pnet_macros, I'd appreciate it if you could add test cases to the relevant directories in pnet_macros/tests, so we don't end up getting regressions.

mrmonday:
OK, I've taken a look through your changes, just to summarise my comments (and add to them):

 * I would like a test case adding to `pnet_macros/tests/run-pass`, which would fail with the current libpnet, but pass with your changes.
 * There's a couple of little tweaks I'd like just to clarify things.
 * When you're done, could you squash your commits - just to get rid of the extra merge commits

Overall this looks good, thank you for your contributions! I look forward to merging it.

alexsult:
Thanks, I'll post something by the end of the week.

PR #155, opened 2016-01-02
ipv4 length and packet in packet handling (options in ipv4 header)

alexsult:
Hi,

I accidentally removed the PR #153. 

I have a small piece of code where I try to get the options of an IPv4 packet

    header = Ipv4Packet::new(ethernet.payload()).unwrap();
    let options = header.get_options();
    println!("Options {:?}",options);

And when it reads a packet,

(built with scary : 

    ip_packet=IP(src="1.1.1.1", dst="2.2.2.2", options=IPOption('\x83\x03\x10'))

)

I have the following message :

    'slice index starts at 20 but ends at 1', ../src/libcore/slice.rs:518                                                                                                                                                              
So I modified the following files :

In src/packet/ipv4.rs:

I modified ipv4_options_length :
. the function did return the length in "words" instead of returning in bytes
. few ipv4 options are only 3 bytes long (EOL, NOP) and don't contain a length field. The easiest way (I found) to handle this, is to define that field (the length) as a Vector with a length deduced from the option number.
. and I added a test for the ipv4 option packet

In pnet_macros/src/decorator.rs:
. I modified the end of the slices to the "offset + length" instead of "length".
. in pnet_macros/tests/run-pass/packet_in_packet.rs I did add a test which illustrates the problem.

Best regards,

Alexis

### DIFF
--- a/pnet_macros/src/decorator.rs
+++ b/pnet_macros/src/decorator.rs
@@ -597,9 +597,9 @@ fn handle_vector_field(cx: &mut GenContext,
                                 pub fn get_{name}_raw(&self) -> &[u8] {{
                                     let _self = self;
                                     let current_offset = {co};
-                                    let len = {packet_length};
+                                    let end = current_offset + {packet_length};
 
-                                    &_self.packet[current_offset..len]
+                                    &_self.packet[current_offset..end]
                                 }}
                                 ",
                                 accessors = accessors,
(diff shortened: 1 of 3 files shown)


mrmonday:
Does this address the issues I raised in the last  pull request? Is it ready for me to review it?

If so, I'll try and look through it now, and hopefully get it merged before my monster pull request that's likely to break this!

alexsult:
Yes, I did what you asked:
- I did modify the names of the functions
- I add the Ipv4OptionNumbers 
- I did add test which fails with the current version and pass with my modifications on decorator.rs.
It "works on my computer".

mrmonday:
This looks great, thank you! I'll merge it once AppVeyor reports success :)

mrmonday:
@homu r+

Thanks for your work!

PR #162, opened 2016-01-24
Begin support for supporting multiple DLTs

mrmonday:
Wraps existing traits in a macro, creating one for each DLT. Ideally this
would be done using generics, however after multiple attempts to get this
working I have not succeeded.

See also issue #1.

### DIFF
--- a/LICENSE-MIT
+++ b/LICENSE-MIT
@@ -1,4 +1,4 @@
-Copyright (c) 2014 Robert Clipsham
+Copyright (c) 2014-2016 Robert Clipsham
 
 Permission is hereby granted, free of charge, to any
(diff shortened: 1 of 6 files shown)


mrmonday:
@homu r+

mrmonday:
For the curious, this is the first in a series of changes I'll be making to support other DLTs. The next step will be adding support for per-backend configuration, at which point I will add in the mechanism for actually supporting different DLTs.

Additionally, if anyone happens to know how to do this with generics rather than macros, I'd love to switch over - I've had very little success with that though (I've managed to get a few variants compiling, but then they haven't actually been usable...)

PR #163, opened 2016-02-02
update to libc 0.2.x

aleksander:
I want my crate only depends on one libc version (and the only reverse dependency for libc 0.1.x is pnet) so I update it. I'm not sure I did this right. The code is "stolen" from nix-rust crate (https://github.com/carllerche/nix-rust/blob/master/src/sys/socket/addr.rs#L275-L331). Please comment.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -32,7 +32,7 @@ optional = true
 version = ">=0.0"
 
 [dependencies.libc]
-version = "=0.1.12"
+version = "0.2.*"
 
 [dependencies.syntex]
 version = "=0.26.0"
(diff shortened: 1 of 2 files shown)


mrmonday:
Hi there,

Sorry I missed your pull request! As you've noticed, upgrading libc is a lot more involved for libpnet, since it needs to support Windows as well, and a lot of the functionality in libc 0.1 has been moved to the winapi crate in libc 0.2 and above.

If you would like some help figuring out how to do this I'd be happy to talk you through the changes, I understand if not though.

aleksander:
Unfortunately, I don't have windows machines and windows is not a target for me. It's better for this to live in a fork I think.

PR #165, opened 2016-02-13
Fix OSX IPv4 header byte order

faern:
FreeBSD / OSX require the fragment offset bytes to be in host order before sent off to libc. This fix replaces the non working bit switch with a hopefully working version.

cc @mrmonday


mrmonday:
Fixes issue #165

@homu r+

PR #167, opened 2016-03-01
Make simple types hashable

faern:
I needed to store EtherType as a key in a HashMap, then I realized that a lot of these quite simple types could make use of being hashable.

@mrmonday


mrmonday:
Looks good, thanks!

@homu r+

PR #170, opened 2016-03-05
Update syntex

mrmonday:
Update syntex to the latest version. Should also fix build on Rust nightly.

Fixed a couple of clippy warnings.


mrmonday:
@homu r+

Issue #1, opened 2014-08-31
Support alternative DLTs


mrmonday:
Currently `DataLinkStream` is hard coded to use Ethernet - it should support arbitrary data link types.

mrmonday:
Closed by #184

PR #171, opened 2016-03-06
Handle headers on loopback devices on bsd/osx

faern:
The previous version prepended 4 bytes == AF_INET to datalink sends. On my computer this header is added automatically so it doesn't seem to be needed. I didn't find any documentation describing this and I only tested it on my OS X 10.11.3 laptop, so I can't guarantee it works for every macos/freebsd flavor.

The previous version also kept the ethernet header. I don't think these are supposed to be included on the loopback interface. This PR changes so when sending to loopback on the datalink layer we discard the header and simply send the payload.

During read the OS will only give us the payload (prepended by the 4 bytes AF_INET) but the API force us to return an `EthernetPacket`. This PR changes so that the 14 bytes before the payload is zeroed out and included in the returned value. Thus, all packets read from a loopback device will have zeroes in all ethernet header fields.

@mrmonday

### DIFF
--- a/src/datalink/bpf.rs
+++ b/src/datalink/bpf.rs
@@ -9,7 +9,6 @@
 extern crate libc;
 
 use std::collections::VecDeque;
-use std::cmp;
 use std::ffi::CString;
 use std::io;
 use std::iter::repeat;
@@ -122,8 +121,6 @@ pub fn datalink_channel(network_interface: &NetworkInterface,
         return Err(err);
     }
 
-    let mut header_size = 0;
-
     // Get the device type
     let mut dlt: libc::c_uint = 0;
     if unsafe { bpf::ioctl(fd, bpf::BIOCGDLT, &mut dlt) } == -1 {
@@ -134,16 +131,20 @@ pub fn datalink_channel(network_interface: &NetworkInterface,
         return Err(err);
     }
 
+    let mut loopback = false;
+    let mut allocated_read_buffer_size = read_buffer_size;
     // The loopback device does weird things
     // FIXME This should really just be another L2 packet header type
     if dlt == bpf::DLT_NULL {
-        header_size = 4;
+        loopback = true;
+        // So we can guaranatee that we can have a header before the packet.
+        // Loopback packets arrive without the header.
+        allocated_read_buffer_size += EthernetPacket::minimum_packet_size();
 
         // Allow packets to be read back after they are written
         if let Err(e) = set_feedback(fd) {
             return Err(e);
         }
-
     } else {
         // Don't fill in source MAC
         if unsafe { bpf::ioctl(fd, bpf::BIOCSHDRCMPLT, &one) } == -1 {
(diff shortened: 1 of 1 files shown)


src/datalink/bpf.rs
```diff
@@ -253,30 +261,43 @@ pub struct DataLinkChannelIteratorImpl<'a> {
...
                 }
             }
         }
-        let (start, len) = self.packets.pop_front().unwrap();
+        let (start, mut len) = self.packets.pop_front().unwrap();
+        len += buffer_offset;
+        // Zero out part that will become fake ethernet header if on loopback.
+        for i in (&mut self.pc.read_buffer[start..start + buffer_offset]).iter_mut() {
+            *i = 0;
+        }
```
faern: Is there any shorter way to zero out a slice? It feels like there should be a nice functional style, one-liner laying around that I'm missing :)

src/datalink/bpf.rs
```diff
@@ -27,11 +26,12 @@ use util::NetworkInterface;
 
 // NOTE buffer must be word aligned.
 #[inline]
-pub fn datalink_channel(network_interface: &NetworkInterface,
-                        write_buffer_size: usize,
-                        read_buffer_size: usize,
-                        channel_type: DataLinkChannelType)
-    -> io::Result<(Box<EthernetDataLinkSender>, Box<EthernetDataLinkReceiver>)> {
+pub fn datalink_channel
+    (network_interface: &NetworkInterface,
```
mrmonday: Why this change? I don't like the arguments starting on the next line. I assume this was rustfmt's work?

src/datalink/bpf.rs
mrmonday: I agree, but I don't know of any way, off the top of my head.

mrmonday:
Just to repeat what I said on IRC:

I don't like this, since it's fixing a hack which shouldn't be there in the first place.

However, you make a convincing point that it currently doesn't work at all, and this makes it sort-of work (much like the hack was intended to).

Other than the stylistic change, I'm happy (perhaps the wrong word :D) to pull it, with the hope that we can do it properly at a later point in time.

src/datalink/bpf.rs
faern: Yes it is. I obey rustfmt all the time. Actually my editor automatically runs it on every save, so this just ended up here. I could back it out but before doing that I would like to ask you to consider going full rustfmt on the entire project instead. It's so nice to remove the human aspect of code style and make it fully automatic. I don't love every line rustfmt gives me, but at least it will be consistent and I don't have to argue with anyone about it because I didn't make the decision :)

faern:
Do you have any suggestion on how this proper solution would look? I think this is pretty OK right now since all information we have access to is the payload, there are no headers but still a datalink packet. I'm not sure what this protocol is named, should there be a different datalink packet type for this?

src/datalink/bpf.rs
mrmonday: I'm sure there's an option to stop it doing this in there - could you investigate? There's already a rustfmt configuration file in the repo.

I would love to start using rustfmt for everything, but unfortunately it's not currently possible (correct me if I'm wrong/this has changed) to exclude areas. Specifically, see https://github.com/libpnet/libpnet/blob/f08e08e76e6dc7498171982f58caa7b2c1270b89/src/packet/ipv4.rs.in#L263-L272 - last time I checked, rustfmt absolutely destroyed the formatting on that array literal, making it completely illegible.

If it doesn't do that any more, I'd be happy to start using it, and perhaps integrating it with the CI.

mrmonday:
The proper solution is to fix issue #1, which I've been gradually chipping away at - there would be an alternative packet type for it (NullPacket, I imagine) which would be used instead.

src/datalink/bpf.rs
faern: Yes, it still destroys that special format. I'm divided here. Configuring the formatter brings back the human factor :( It would be so beautiful to remove that completely.
It's possible to add `#[cfg_attr(rustfmt, rustfmt_skip)]` to make it skip formating, but I only get that to work for entire files, not individual lines or functions.

I'll simply back out the reformatting.

mrmonday:
@homu r+

PR #173, opened 2016-03-20
Use std::net::IpAddr instead of own impl

faern:
Since `std::net::IpAddr` has been stabilized since rustc 1.7.0 we might want to use the std implementation instead. That enum has been around as unstable since 1.0.0, was deprecated in 1.6.0 and then brought back to stable in 1.7.0

@mrmonday

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,7 +3,7 @@ cache:
     directories:
         - $HOME/.cargo
 rust:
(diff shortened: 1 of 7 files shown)


mrmonday:
@homu r+

mrmonday:
@homu r+

PR #174, opened 2016-03-25
Split internal per target

NSil:
Split the 'internal' module to unix and windows sub-modules.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,3 @@
 target/
 Cargo.lock
-Packet.lib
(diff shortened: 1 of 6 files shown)


.gitignore
```diff
@@ -1,2 +1,3 @@
 target/
 Cargo.lock
+Packet.lib
```
mrmonday: Could you make this `*.lib` instead please?

mrmonday:
@homu r+

PR #175, opened 2016-03-26
Fix a bug where for Windows, the datalink channel's read and write bu…

NSil:
…ffer sizes were flipped.


mrmonday:
Good catch, thanks!

@homu r+

PR #176, opened 2016-03-30
Proposal to make libpnet compliant with RFC #1445

wictory:
https://github.com/rust-lang/rfcs/pull/1445

It seems that this RFC breaks libpnet, and it didn't compile with the newest nightly so maybe it's time to act.

### DIFF
--- a/benches/rs_sender.rs
+++ b/benches/rs_sender.rs
@@ -11,7 +11,7 @@ extern crate pnet;
 use pnet::datalink::datalink_channel;
 use pnet::datalink::DataLinkChannelType::Layer2;
 use pnet::packet::{MutablePacket, Packet};
-use pnet::packet::ethernet::{EtherTypes, MutableEthernetPacket, EthernetPacket};
+use pnet::packet::ethernet::{EtherType, EtherTypes, MutableEthernetPacket, EthernetPacket};
 use pnet::packet::ip::IpNextHeaderProtocols;
 use pnet::packet::ipv4::MutableIpv4Packet;
 use pnet::packet::ipv4;
@@ -97,7 +97,7 @@ fn main() {
     {
         mut_ethernet_header.set_destination(destination);
         mut_ethernet_header.set_source(interface.mac_address());
-        mut_ethernet_header.set_ethertype(EtherTypes::Ipv4);
+        mut_ethernet_header.set_ethertype(EtherType::new(EtherTypes::Ipv4));
         build_udp4_packet(mut_ethernet_header.payload_mut(), "rmesg");
     }
 
--- a/examples/packetdump.rs
+++ b/examples/packetdump.rs
@@ -100,7 +100,7 @@ fn handle_arp_packet(interface_name: &str, ethernet: &EthernetPacket) {
 }
 
 fn handle_packet(interface_name: &str, ethernet: &EthernetPacket) {
-    match ethernet.get_ethertype() {
+    match ethernet.get_ethertype().unpack() {
         EtherTypes::Ipv4 => handle_ipv4_packet(interface_name, ethernet),
         EtherTypes::Ipv6 => handle_ipv6_packet(interface_name, ethernet),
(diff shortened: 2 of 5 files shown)


mrmonday:
Hi there!

Thank you for your pull request.

I haven't had time to look through the RFC - what is the issue? I really don't like this solution, since it removes a lot of type safety (those constants are newtypes for a reason - they're not just random integers). Is there another way to solve this?

wictory:
Hi!

Thanks for your comment. 

As the matter of fact I also think that type safety is lost, and I think now that my PR might be a bit unconsidered. The solution I proposed was to change the Ipv4OptionNumber and EtherType constants into u8:s and u16:s. In the RFC discussion it is mentioned that a change from `match foo { BAR => ... }` to `match foo { c if c == BAR => ... }` would be possible. This change would preserve the type safety while introducing only a little more complexity.

If people think this is a good idea I can put together a new PR.

mrmonday:
Note that there's also the [IpNextHeaderProtocols](https://github.com/libpnet/libpnet/blob/85c6a36696d20114298e4fc22ca0f1b0659ad72b/src/packet/ip.rs#L23) that this will affect (there are possibly other things too - newtypes are a fairly common way of dealing with constants in libpnet, since they add type-safety, without restricting you to a particular set of values like an `enum` would).

It's a shame to have the additional verbosity - is there any other choice? Perhaps we could implement [Into](https://doc.rust-lang.org/std/convert/trait.Into.html) for these, then use `match foo.into() { const1.into() => ... }` etc?

I don't know the best course of action here. I'm really disappointed that this RFC has been implemented, since it breaks backwards compatibility for (as far as I can tell - I haven't had time to read the RFC in full) negative gain. What happened to semver? It looks like it's in FCP, perhaps it's worth commenting about this, before it is accepted?

wictory:
I find it strange that IpNextHeaderProtocols goes through the compiler. I can't find a reason why it should.

Another way to deal with the matching would be to have bidirectional mapping from the newtype (or the integer) to an enum which is matchable, so that the constant definition would be inside a kind of function mapping integers to enums. Maybe a macro would be needed to create this mapping properly, in particular if this pattern is used all over the library.

As the matter of fact I wrote a macro for it now. The complete "type" for EtherType would be:
```
bidir_mapping!{
    EtherType(u16) -> EtherTypes :
    Ipv4 = 0x0800,
    Arp  = 0x0806,
    WakeOnLan = 0x0842,
    Rarp = 0x8035,
    Ipv6 = 0x86DD;
    Unknown
}

impl EtherType {
    /// Construct a new EtherType
    pub fn new(val: u16) -> EtherType {
        EtherType(val)
    }
}
```

The macro creates the newtype and an enum type, which has implemented `From<$enumtype> for $newtype` and `From<$newtype> for $enumtype`. One only needs to insert a couple of .into():s into the code. If you are interested I can make a pull request. It might also make sense to make a completely new crate for the macro also if the RFC goes through as it is and other people are pissed off.

mrmonday:
There's nothing wrong with the IpNextHeaderProtocols - it's likely that nothing matches on them, so there are no compile errors related to it.

I've left a comment on the RFC asking about the best course of action, hopefully there'll be a useful response.

mrmonday:
Looks like there shouldn't need to be any code changes - there's a compiler bug, and a syntex issue at play here -  https://github.com/rust-lang/rust/issues/32655. With any luck, this should resolve itself in a future nightly.

wictory:
How interesting, well hopefully it gets resolved by that.

wictory:
This explains why the complier did not complain about [IpNextHeaderProtocols](https://github.com/libpnet/libpnet/blob/85c6a36696d20114298e4fc22ca0f1b0659ad72b/src/packet/ip.rs#L23).

mrmonday:
I'm gonna close this, since upgrading syntex has allowed libpnet to start working on nightly again - https://github.com/rust-lang/rfcs/issues/325#issuecomment-56926977.

Feel free to reopen it (or ask me to), if you believe this was a mistake.

Thank you for your time!

PR #181, opened 2016-04-17
Use slices where possible as we are copying anyway

polachok:
Why do we use Vec's here and then copy them?


mrmonday:
I suppose there was a reason at the time, but this is clearly the way it should be done.

Thank you for all your work recently! You're making a lot of really valuable contributions.

@homu r+

PR #183, opened 2016-04-22
Provide iterator interface for Vec fields

polachok:


mrmonday:
Looks good, thank you!

@homu r+

PR #185, opened 2016-05-17
Support inner modules for `#[packet]`

mrmonday:
Previously, submodules in crates were not recursed into when removing
attributes, causing compilation errors when `#[packet]`s are found in inner
modules.


mrmonday:
@homu r+

PR #193, opened 2016-05-31
Implemented read and write timeouts

moosingin3space:
WinPcap is not supported here.

Needs unit tests and other platform backends.
- [x] Linux backend
- [x] BPF backend
- [x] Netmap backend
- [x] Cross-platform unit tests

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,3 +8,59 @@ Cargo.lock
 *.tar.gz
 *.patch
 *.txt
+
+### Emacs ###
+# -*- mode: gitignore; -*-
+*~
+\#*\#
+/.emacs.desktop
+/.emacs.desktop.lock
+*.elc
+auto-save-list
+tramp
+.\#*
+
+# Org-mode
+.org-id-locations
+*_archive
+
+# flymake-mode
+*_flymake.*
+
+# eshell files
+/eshell/history
+/eshell/lastdir
+
+# elpa packages
+/elpa/
+
+# reftex files
+*.rel
+
+# AUCTeX auto folder
+/auto/
+
+# cask packages
+.cask/
+dist/
+
+# Flycheck
+flycheck_*.el
+
+# server auth directory
+/server/
+
+# projectiles files
+.projectile
+
+### Vim ###
+# swap
+[._]*.s[a-w][a-z]
+[._]s[a-w][a-z]
+# session
+Session.vim
+# temporary
+.netrwhist
+*~
+# auto-generated tag files
+tags
--- a/.gitignore
+++ b/.gitignore
@@ -58,7 +58,7 @@ flycheck_*.el
 [._]*.s[a-w][a-z]
 [._]s[a-w][a-z]
 # session
-Session.vim
+*.vim
 # temporary
 .netrwhist
 *~
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "pnet"
-version = "0.11.1"
+version = "0.12.0"
 authors = [ "Robert Clipsham <robert@octarineparrot.com>" ]
 license = "MIT/Apache-2.0"
 homepage = "https://github.com/libpnet/libpnet"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -32,7 +32,13 @@ optional = true
 version = ">=0.0"
 
 [dependencies.libc]
-version = "=0.1.12"
+version = "=0.2.11"
+
+[dependencies.winapi]
+version = "=0.2.7"
+
+[dependencies.ws2_32-sys]
+version = "=0.2.1"
 
 [dependencies.syntex]
 version = "0.31.*"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -32,13 +32,13 @@ optional = true
 version = ">=0.0"
 
 [dependencies.libc]
-version = "=0.2.11"
+version = "0.2.*"
 
 [dependencies.winapi]
-version = "=0.2.7"
+version = "0.2.*"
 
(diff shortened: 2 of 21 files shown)


moosingin3space:
Note: I'd like to make the `duration_to_timeval` function public

moosingin3space:
Wow, sorry, clicked the wrong button.

src/internal/mod.rs
```diff
@@ -82,3 +83,10 @@ pub fn recv_from(socket: CSocket,
         Ok(len as usize)
     }
 }
+
+pub fn duration_to_timeval(dur: Duration) -> libc::timeval {
```
mrmonday: This function doesn't build on Windows due to type mismatches - check the build output for AppVeyor to see what's going on.

mrmonday:
This looks great so far! Let me know when you've got something you're happy to be merged.

How much work are you planning on doing in terms of platform support? If you need assistance getting things to work on other platforms I'm happy to help. The Travis/AppVeyor should let you feel your way to a solution if you don't have access to the other supported platforms.

I'm happy to merge this without complete platform support*, but I think it should have support for at least two different systems so that we know it definitely is somewhat portable.

\* Obviously we ideally want to support everything, I understand that it's not always simple/straight forwards to do so though.

moosingin3space:
I regularly use Linux and FreeBSD machines, so I could probably get BPF and *maybe* Netmap support (I do not have any supported hardware for Netmap). I have access to a Windows 10 machine, but it isn't set up for development, so it would be some time before I could have the WinPcap backend working.

moosingin3space:
The unit test is unfortunately pretty bad at actually testing timeouts. I would like some help regarding how best to test this timeout functionality.

mrmonday:
Looks like BPF and WinPcap aren't building, and the tests are failing on Linux?

src/test.rs
```diff
@@ -394,6 +395,70 @@ fn layer2() {
...
+    let (tx, rx) = channel();
+
+    let cfg = datalink::Config {
+        write_buffer_size: 4096,
+        read_buffer_size: 4096,
+        channel_type: datalink::ChannelType::Layer2,
+        bpf_fd_attempts: 100,
+        read_timeout: Some(Duration::from_millis(300)),
+        write_timeout: Some(Duration::from_millis(100))
+    };
```
mrmonday: Rather than filling in all these fields manually, you can use [record update syntax](https://doc.rust-lang.org/book/structs.html#update-syntax), eg:
```rust
SomeStruct {
    field1: value,
    ..Default::default()
}
```
This saves filling out all the irrelevant defaults, and makes it clear which fields you care about.

src/test.rs
```diff
@@ -394,6 +395,70 @@ fn layer2() {
...
+                Err(e) => {
+                    assert!(e.kind() == ErrorKind::WouldBlock);
+                }
+            }
+        }
+    });
+    rx.recv().unwrap();
+    // Wait a while
+    thread::sleep(Duration::from_millis(1000));
+    // don't even send anything
```
mrmonday: There are a couple of options here, but they're ultimately just workarounds until we start testing against a dedicated virtual network where we can control the traffic:
 * Decrease the timeouts to reduce the chance of stray traffic interfering
 * Retry the test a few times to see if it passes
There's probably something else we could do too - I guess we wait and see how flaky the test is.

moosingin3space: If Travis uses a recent-enough (works for me on Ubuntu 14.04) Linux, we could have Travis create an `iproute2` dummy interface for testing.

moosingin3space:
A side note - I figured out a way to use the latest `libc` crate while investigating problems. No regressions were introduced, but I did have to write a few more lines of `unsafe` code.

src/test.rs
moosingin3space: I discovered how to make this work -- it requires two sockets to be created -- one for sending and another for receiving. Additionally, the dummy interface must have an IP address:
```sh
sudo ip addr add 192.168.0.50 dummy0
```

Would you like me to introduce this to the Linux tests on Travis?

moosingin3space:
Travis on Linux seems to be ignoring unit tests.

moosingin3space:
How do the unit tests choose an interface to do testing on OS X?

Cargo.toml
```diff
@@ -32,7 +32,13 @@ optional = true
 version = ">=0.0"
 
 [dependencies.libc]
-version = "=0.1.12"
+version = "=0.2.11"
```
mrmonday: Can you make this, and the dependencies below `0.2.*` etc, rather than pinning them to the specific version please?

.gitignore
```diff
@@ -8,3 +8,59 @@ Cargo.lock
...
+
+# projectiles files
+.projectile
+
+### Vim ###
+# swap
+[._]*.s[a-w][a-z]
+[._]s[a-w][a-z]
+# session
+Session.vim
```
mrmonday: May as well make this `*.vim`.

src/datalink/netmap.rs
```diff
@@ -94,39 +95,70 @@ impl Drop for NmDesc {
...
+            tv_nsec: to.subsec_nanos() as libc::c_long
+        } as *const libc::timespec,
+        None => ptr::null()
     }
 }
 
 /// Create a datalink channel using the netmap library
 #[inline]
-pub fn channel(network_interface: &NetworkInterface, _config: Config)
+pub fn channel(network_interface: &NetworkInterface, config: &Config)
```
mrmonday: I believe this is a merge conflict? Should be `Config` now.

mrmonday:
Great stuff, thank you for your continued work on this!

@homu r+

PR #194, opened 2016-06-01
WIP: Implemented incoming packet timestamps

moosingin3space:
Right now, as with #193, this is Linux-only, but I do not know how to support other platforms.

Needs unit tests and a discussion regarding the design of `TimestampedEthernetPacket`.

### DIFF
--- a/src/datalink/mod.rs
+++ b/src/datalink/mod.rs
@@ -65,6 +65,9 @@ pub enum ChannelType {
     Layer3(EtherType),
 }
 
+/// Type of timestamped Ethernet packet (Linux only)
+pub struct TimestampedEthernetPacket<'a>(Duration, EthernetPacket<'a>);
+
 /// A channel for sending and receiving at the data link layer
 ///
 /// NOTE: It is important to always include a catch-all variant in match statements using this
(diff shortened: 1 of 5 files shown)


moosingin3space:
There's an efficiency improvement I can gain by using `recvmsg` instead of `recvfrom` when hardware timestamps are enabled. Will investigate.

src/datalink/mod.rs
```diff
@@ -58,6 +59,9 @@ pub enum ChannelType {
     Layer3(EtherType),
 }
 
+/// Type of timestamped Ethernet packet (Linux only)
+pub struct TimestampedEthernetPacket<'a>(Duration, EthernetPacket<'a>);
```
mrmonday: This name is confusing, since this isn't actually a `#[packet]`. I think it would probably be better to just use a tuple instead.

mrmonday:
Left a small comment here, see also my comments in #193.

I'll do a more thorough review when you're ready for it!

Thanks for your work!

src/datalink/mod.rs
moosingin3space: The macro raised errors when I used a tuple. Or are you suggesting `type EthernetPacketTimestamped<'a> = (Duration, EthernetPacket<'a>)`, which can then be used with `let (ts, pkt) = iter.next()`?

mrmonday:
Hi @moosingin3space - are you still working on this?

What is the status of it?

mrmonday:
Closing due to inactivity - give me a shout if you pick this up again!

PR #203, opened 2016-07-07
support Vec<u{16,24,32,..}> in packets

ewust:
Currently, `Vec<T>` is only allowed in a #[packet] struct if `T` is u8 or another #[packet] struct.

This adds support for when `T` is a byte-aligned big-endian primitive (e.g. u16be, u24be, u32be, u40be, u48be, u56be, and u64be). This is useful if you have an array of 2-byte values in a packet header (such as in TLS cipher suites, or elliptic curves).

The GRE packet definition would also benefit from this, as it is currently using a #[packet] definition U16BE and U32BE, which carry extra weight with unused payloads, and prevent those fields from having easily-defined constants (since you can't instantiate the payload Vec<u8> at compile time).

### DIFF
--- a/pnet_macros/src/decorator.rs
+++ b/pnet_macros/src/decorator.rs
@@ -1173,6 +1173,37 @@ fn generate_accessor_op_str(name: &str,
     op_strings
 }
 
+#[test]
+fn test_generate_accessor_op_str() {
+
+    {
+        let ops = operations(0, 24).unwrap();
+        let result = generate_accessor_op_str("test", "u24be", &ops);
+        let expected = "let b0 = ((test[co + 0] as u24be) << 16) as u24be;\n\
+                    let b1 = ((test[co + 1] as u24be) << 8) as u24be;\n\
+                    let b2 = ((test[co + 2] as u24be)) as u24be;\n\n\
+                    b0 | b1 | b2\n";
+
+        assert_eq!(result, expected);
+    }
+
+    {
+        let ops = operations(0, 16).unwrap();
+        let result = generate_accessor_op_str("test", "u16be", &ops);
+        let expected = "let b0 = ((test[co + 0] as u16be) << 8) as u16be;\n\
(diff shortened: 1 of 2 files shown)


pnet_macros/src/decorator.rs
```diff
@@ -534,16 +537,25 @@ fn handle_vec_primitive(cx: &mut GenContext,
                                         let end = current_offset + {packet_length};
 
                                         let packet = &_self.packet[current_offset..end];
-                                        let mut vec = Vec::with_capacity(packet.len());
-                                        vec.extend_from_slice(packet);
+                                        let mut vec : Vec<{inner_ty_str}> = Vec::with_capacity(packet.len());
```
mrmonday: No space before the colon after `let mut vec` please :)

pnet_macros/src/decorator.rs
```diff
@@ -534,16 +537,25 @@ fn handle_vec_primitive(cx: &mut GenContext,
...
+                                        let mut co = 0;
+                                        for _ in 0..vec.capacity() {{
+                                            vec.push({{
+                                                {ops}
+                                            }});
+                                            co += {size};
+                                        }}
+                                        //vec.extend_from_slice(packet);
```
mrmonday: Could you remove this commented out line please?

mrmonday:
Thank you for your pull request!

This looks good, other than the minor stylistic changes I've noted. Could you please add a couple of test cases for this? The `pnet_macros` code is fairly fragile, and everything else depends on it. Thanks!

ewust:
Thanks, I made the simple style changes as requested. Happy to make unit tests, but having trouble compiling running the existing tests for pnet_macros (compiletest_rs doesn't like any stable/beta rustc version, and using the current nightly throws errors due to syntax::ast changes).

Will try with an older nightly build, but suggestions welcome. :)

mrmonday:
Looking good, thanks!

@homu r+

PR #204, opened 2016-07-08
Move get_network_interfaces into the datalink mod

faern:
Different backends might have different interfaces. So I moved `util::get_network_interfaces` to `datalink::interfaces` and the implementations to the different backends.

@mrmonday


mrmonday:
Some quick comments, since this is WIP:

 * I would rather use the name "interfaces" than "ifaces". Perhaps "unix" would be even better, since that's really what it's for?
 * I think `sockaddr_to_network_addr()` should probably still exist in util rather than under datalink
 * I like the rename to `interfaces()` rather than `get_network_interfaces()`
 * The documentation will need updating too

Looks like a solid start to this, good work! :)

faern:
* I fully agree to the rename. Now it's `unix_interfaces.rs`
* `sockaddr_to_network_addr` might be a utility function but it's not part of the public API and it's only used by the unix verison of interfaces. Also if it's kept in util we need to make it public to be able to use it ourselves.

faern:
You said something about not having `interfaces` public. Why should it not be public? It's public before this PR. Also I think `datalink` should to expose the same API as all the backends beneath it.

faern:
@mrmonday I'm thinking about moving `util::NetworkInterface` to `datalink::NetworkInterface`. What do you think about that?

faern:
@mrmonday `#[deprecated]` was stabilized in Rust 1.9, so automatic tests fail since they run on 1.7. What do you think is the correct solution? Not use `#[deprecated]` or only support Rust >=1.9.

According to the large survey they conducted quite recently*, almost no one is using anything older than the latest release. But that is totally up to you to decide.

* https://blog.rust-lang.org/2016/06/30/State-of-Rust-Survey-2016.html

faern:
To not make the `datalink` module root dirty I could put this new stuff under a submodule `datalink::util`
* `datalink::unix_interfaces` becomes `datalink::util::unix`
* `util::NetworkInterface` becomes `datalink::util::NetworkInterface` (If you like that move at all, see earlier question)

mrmonday:
> `#[deprecated]` was stabilized in Rust 1.9
Modify the `.travis.yml` and `appveyor.yml` files and bump the minimum required versions. I don't know if we mention a minimum Rust version in the README, but if we do, update that too.

`datalink::NetworkInterface` works for me. I don't see any need for a `util` namespace under datalink, unless we introduce utilities which are not directly related to the datalink but still useful.

src/util.rs
```diff
@@ -153,231 +148,13 @@ impl NetworkInterface {
...
-                Err(_) => (None, None),
-            }
-        }
-    }
-}
-
 /// Get a list of available network interfaces for the current machine.
+/// Deprecated. Instead use the implementation available for your backend.
+/// The default one is at `pnet::datalink::interfaces`.
+#[deprecated]
```
mrmonday: Does `#[deprecated]` accept a message parameter? It would be nice to mention it has moved to `pnet::datalink::interfaces()` if it does.

PR #205, opened 2016-07-11
Add dummy datalink backend for testing

faern:
Adding a datalink backend for testing. It's based on in memory FIFO queues that can be used to inject and read packets on the fake network. Useful for testing stuff built on top of pnet.

@mrmonday

### DIFF
--- a/src/datalink/dummy.rs
+++ b/src/datalink/dummy.rs
@@ -24,11 +24,11 @@ use util::MacAddr;
 /// channels that are used to communicate with the fake network.
 #[derive(Debug)]
 pub struct Config {
-    in_packets_rx: Receiver<io::Result<Box<[u8]>>>,
-    in_packets_tx: Option<Sender<io::Result<Box<[u8]>>>>,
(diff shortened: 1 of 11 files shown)


src/datalink/dummy.rs
```diff
@@ -0,0 +1,349 @@
...
+    use packet::{MutablePacket, Packet};
+    use packet::ethernet::{EthernetPacket, MutableEthernetPacket};
+    use datalink::{EthernetDataLinkReceiver, EthernetDataLinkSender};
+    use datalink::Channel::Ethernet;
+
+    #[test]
+    fn send_too_small_packet_size() {
+        let (_, _, mut tx, _) = create_net();
+        // Check that it fails to send with too small packet sizes
+        assert!(tx.build_and_send(1, 0, &mut |_| {}).is_none());
```
mrmonday: This raises an interesting question - I think it should be possible to send packets which aren't valid Ethernet packets (because they're too small). I think that's a discussion for another time though.

mrmonday:
@homu r+

PR #206, opened 2016-07-12
Version bump

mrmonday:
Also ignore a few more files.


mrmonday:
@homu r+

faern:
@mrmonday We probably want to ignore `*.bk` which are the backup files generated by `rustfmt`

PR #207, opened 2016-07-12
Make all types inside IcmpPacket hashable

faern:
I can't see a reason not to have them hashable. I need some of them hashable in [librips](https://github.com/faern/librips)

@mrmonday


mrmonday:
Could you do this for the other newtypes in other modules too please?

mrmonday:
@homu r+

PR #208, opened 2016-07-13
Little endian?

polachok:
I just started working on [Netlink](https://en.wikipedia.org/wiki/Netlink) library based on libpnet/pnet_macros. The problem is that Netlink messages are encoded in little endian.
It seems like this little diff works as intended.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "pnet"
-version = "0.15.0"
+version = "0.16.0"
 authors = [ "Robert Clipsham <robert@octarineparrot.com>" ]
 license = "MIT/Apache-2.0"
 homepage = "https://github.com/libpnet/libpnet"
--- a/README.md
+++ b/README.md
@@ -70,7 +70,7 @@ To use `libpnet` in your project, add the following to your Cargo.toml:
 
 ```
 [dependencies.pnet]
-version = "0.15.0"
+version = "0.16.0"
 ```
 
 `libpnet` should work on any Rust channel (stable, beta, or nightly), starting
--- a/pnet_macros/Cargo.toml
+++ b/pnet_macros/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "pnet_macros"
-version = "0.12.0"
+version = "0.13.0"
 authors = ["Robert Clipsham <robert@octarineparrot.com>"]
 license = "MIT/Apache-2.0"
 homepage = "https://github.com/libpnet/libpnet"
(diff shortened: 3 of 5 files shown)


mrmonday:
Thanks for working on this... I'd really like to see some test cases for this before merging it. All the bit manipulation stuff is hard to get right.

I'm not sure that this will work as expected for eg. u17.

polachok:
You're right, it doesn't work! I'll look into it.

mrmonday:
Are you still working on this @polachok?

rbtcollins:
There aren't any new commits in his branch :( - I'd love to see what he has so far in mainline, to make dependency graphs saner. ..

polachok:
I'll try to find some time to work on this

rbtcollins:
So, month and a bit later: can I minimally get some guidance on the tests needed :- but ideally while the patch may be incomplete for u17 etc, it does work for u8/16/32...

Perhaps separately, netlink actually needs a native endian option; is that something pnet would be interested in supporting?

mrmonday:
I would be willing to accept a patch that only works for power of two field types, providing it fails at compile time for non power of two fields.

By native endianness, do you mean host endianness? I'm weary about supporting this, since it clearly isn't cross-platform - if you can provide a compelling use case then patches are welcome! :)

rbtcollins:
Thanks. The Linux netlink protocol is defined in host endian types but in
Ashok other ways behaves like a network protocol.

mrmonday:
Sounds like a good enough reason to me! Let me know if you need any advise, presuming you want to implement it - I'm happy to mentor this.

mrmonday:
@polachok I'd quite like to get this merged in so that packets are usable outside libpnet again. Before I do that, could you:

 * Add an assertion to check that the argument for the little endian version of the code is a power of two (I assume that is what this works for?)
 * Bump the version numbers for libpnet/libpnet_macros in Cargo.toml and the README

Given that the test suite doesn't work on nightly or stable right now, I'll forgo asking for you to write tests.

With those changes done I'll get this merged/released ASAP.

polachok:
I made a test here:

https://github.com/polachok/libpnet-custom-packet/blob/little-endian/src/packet/test.rs

It seems to work for u15 in this case.

mrmonday:
Great stuff, thank you!

I'll try and release this tonight.

@homu r+

mrmonday:
Released:

 * https://crates.io/crates/pnet_macros/0.13.0
 * https://crates.io/crates/pnet/0.16.0

PR #212, opened 2016-08-23
Update unix_interfaces with CStr compatibility

rschifflin:
Ran into some issues cross-compiling on architectures where `c_char` is conditionally defined as `u8` rather than `i8`, causing `CStr::from_ptr` to reject the hard `u8` cast. Updated the cast to `c_char` directly


mrmonday:
Great stuff, thanks!

@homu r+

PR #215, opened 2016-08-25
Checksum improvement

faern:
According to what I have found the checksums of both ipv4 and icmp packets should be calculated as if the checksum field is zero. Today both the functions use the checksum field as is in the calculation and will thus return the wrong checksum when the packet contains a non-zero checksum.

For outgoing packets this is easy to work around by just setting the checksum to zero before doing the calculation. But when receiving incoming packets in the non-mutable form it's not possible to validate the checksum of that packet.

https://en.wikipedia.org/wiki/IPv4#Header_Checksum
https://en.wikipedia.org/wiki/Internet_Control_Message_Protocol#header_checksum

This PR changes so both functions skip over the checksum word in the headers. This PR also does some other small code cleanups of those functions as well as add tests. The tests does not pass with the old implementation, but with the new.

@mrmonday


mrmonday:
This looks great, thank you!

@homu r+

faern:
I added new commits that adds benchmarks and moves the checksum implementation to the util module.

Benchmarking results before any refactor:
```
test packet::icmp::checksum_tests::bench_checksum_large ... bench:         576 ns/iter (+/- 57) 
test packet::icmp::checksum_tests::bench_checksum_small ... bench:          12 ns/iter (+/- 6)
test packet::ipv4::checksum_tests::bench_checksum_large ... bench:          11 ns/iter (+/- 2)
test packet::ipv4::checksum_tests::bench_checksum_small ... bench:          12 ns/iter (+/- 5)
```

With the code as proposed in this PR:
```
test packet::icmp::checksum_tests::bench_checksum_large ... bench:         365 ns/iter (+/- 27) 
test packet::icmp::checksum_tests::bench_checksum_small ... bench:           8 ns/iter (+/- 1)
test packet::ipv4::checksum_tests::bench_checksum_large ... bench:           7 ns/iter (+/- 1)
test packet::ipv4::checksum_tests::bench_checksum_small ... bench:           7 ns/iter (+/- 2)
```

The changes introduces a bit of `unsafe{}` but it reduces the execution time to ~60% of the original time.

faern:
There is still the problem of checksumming an odd number of bytes. With these last commits at least doing that does not panic, it just ignores the last byte. What is even the correct procedure? How to treat the last byte?

mrmonday:
@homu r+

Great stuff, thanks!

PR #216, opened 2016-09-09
Added padding for Displaying MacAddr

Idleless:


faern:
@Idleless Great! But could you maybe add a testcase? Should be a oneliner more or less.

mrmonday:
Superseded by #222

PR #217, opened 2016-09-15
Allow Packets to own their data

faern:
So this feature is not done yet, I created the PR to show how I'm thinking and get some feedback. I have no idea why it won't build for Icmp. Probably a stupid error I can't spot because it's late. Also, would be nice to clean up the macros and other parts of the `PacketData` stuff.

I want to make it possible for `*Packet` instances to own their own data. The main reason for this is that we can then create functions that return packets with the correct type. For example, imagine these:
- `save_ipv4_fragment_and_fetch_assembled_if_done(pkg: Ipv4Packet<'p>) -> Option<Ipv4Packet<'static>>`
- `generate_test_ipv4() -> Ipv4Packet<'static>`

Which are not possible with today's code.

@mrmonday

### DIFF
--- a/pnet_macros/src/decorator.rs
+++ b/pnet_macros/src/decorator.rs
@@ -946,9 +946,7 @@ fn generate_packet_impl(cx: &mut GenContext,
         #[inline]
         pub fn to_immutable<'p>(&'p self) -> {imm_name}<'p> {{
             use ::pnet::packet::PacketData;
-            match *self {{
-                {name} {{ ref packet }} => {imm_name} {{ packet: PacketData::Borrowed(packet.as_slice()) }}
-            }}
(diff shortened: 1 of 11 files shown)


mrmonday:
Are you still working on this @faern?

faern:
@mrmonday Ah, sorry. I have been traveling for a month, so I didn't put much time into my Rust coding. I just got back home, so I will hopefully have time for rips now again. I think this PR is done, but I have not used it extensively myself yet. In other words, it could be merged and it would work, but I have not verified that the API is in the most usable form yet. However, that could be in a separate PR later as well if we feel the need to improve anything.

mrmonday:
I'll let it sit for now then I think - I'd rather not pull things in which will potentially need changing straight away. Let me know how it works out for you!

faern:
@homu r+

PR #219, opened 2016-09-18
Prevent pselect timeout parameter from being optimized out

### LINKED ISSUE #218: pselect EINVAL when compiling in release mode
I have an application that sets up a channel with a `read_timeout`:
```rust
datalink::Config { read_timeout: Some(Duration::new(10, 0)), ..Default::default() };
```
When compiling my application in debug mode everything works and `strace` shows `pselect` calls with the expected timeout:
```
pselect6(5, [4], NULL, NULL, {10, 0}, {NULL, 8}) = 1 (in [4], left {9, 999998995})
```
But when compiling the same application in release mode, some receive calls fail with an EINVAL error and from the `strace` output it looks like the timeout argument is filled with garbage data:
```
pselect6(5, [4], NULL, NULL, {908191936, 2097152}, {NULL, 8}) = 1 (in [4], left {908191936, 2095630})
pselect6(5, [4], NULL, NULL, {0, 78264}, {NULL, 8}) = 1 (in [4], left {0, 77124})
pselect6(5, [4], NULL, NULL, {0, 2097865011817319249}, {NULL, 8}) = -1 EINVAL (Invalid argument)
pselect6(5, [4], NULL, NULL, {30, 2024323}, {NULL, 8}) = 1 (in [4], left {30, 2021266})
pselect6(5, [4], NULL, NULL, {0, 2097865011889982044}, {NULL, 8}) = -1 EINVAL (Invalid argument)
```

mbudde:
The argument to the map closure may be optimized out by the compiler.
Instead use `as_ref()` to get a reference to the real data.

Fixes #218

### DIFF
--- a/src/datalink/bpf.rs
+++ b/src/datalink/bpf.rs
@@ -276,7 +276,7 @@ impl EthernetDataLinkSender for DataLinkSenderImpl {
                                   ptr::null_mut(),
                                   &mut self.fd_set as *mut libc::fd_set,
                                   ptr::null_mut(),
-                                  self.timeout.map(|to| &to as *const libc::timespec)
+                                  self.timeout.as_ref().map(|to| to as *const libc::timespec)
                                   .unwrap_or(ptr::null()),
                                   ptr::null())
                 };
@@ -317,7 +317,7 @@ impl EthernetDataLinkSender for DataLinkSenderImpl {
                           ptr::null_mut(),
                           &mut self.fd_set as *mut libc::fd_set,
                           ptr::null_mut(),
-                          self.timeout.map(|to| &to as *const libc::timespec)
+                          self.timeout.as_ref().map(|to| to as *const libc::timespec)
                           .unwrap_or(ptr::null()),
                           ptr::null())
         };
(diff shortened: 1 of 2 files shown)


mbudde:
I'm not sure if [`datalink::get_timeout`](https://github.com/libpnet/libpnet/blob/ab6ed4f0f797080db8b2252ba8f3e78487dbf36a/src/datalink/netmap.rs#L127-L130) might have the same problem?

I was able to reproduce the bug with the following program:
```rust
extern crate pnet;

use std::time::Duration;
use pnet::datalink;

fn main() {
    let interface = datalink::interfaces().into_iter()
        .find(|iface| iface.name == "wlan0").unwrap();

    let config =
        datalink::Config { read_timeout: Some(Duration::new(10, 0)), ..Default::default() };
    let (_, mut rx) = match datalink::channel(&interface, config) {
        Ok(datalink::Channel::Ethernet(tx, rx)) => (tx, rx),
        Ok(_) => panic!("unhandled channel type"),
        Err(e) => panic!("unable to create channel: {}", e),
    };

    let mut iter = rx.iter();
    loop {
        match iter.next() {
            Ok(_) => { println!("got packet
            }"); }
            Err(ref e) if e.kind() == ::std::io::ErrorKind::TimedOut => { println!("timed out"); }
            Err(e) => { println!("got error while reading: {}", e); }
        }
    }
}
```

mrmonday:
This looks good, thanks!

`get_timeout()` definitely has a problem - it's returning a pointer to the stack, which definitely won't end well... If you'd like to open another pull request for it I'd be grateful!

@homu r+

PR #220, opened 2016-09-18
Fix timeout in netmap

mbudde:
Change `get_timeout` so it returns a `timespec` struct instead of a pointer to the stack.

Note that I haven't tested that this compiles/works since getting the netmap library seems to be quite a bit more work than an `apt-get`.

References #219


mrmonday:
Looks good to me, thanks!

@homu r+

PR #222, opened 2016-10-06
Implement Display for Ethertype and IpNextHeaderProtocol

fsck:
Also adds leading zeros to Display for MacAddr.

### DIFF
--- a/src/packet/ip.rs
+++ b/src/packet/ip.rs
@@ -10,6 +10,7 @@
 //! fields.
 
 use packet::PrimitiveValues;
+use std::fmt;
 
 /// Protocol numbers as defined at:
 /// http://www.iana.org/assignments/protocol-numbers/protocol-numbers.xhtml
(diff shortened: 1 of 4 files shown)


src/packet/ip.rs
```diff
@@ -481,3 +482,167 @@ impl PrimitiveValues for IpNextHeaderProtocol {
...
+            &IpNextHeaderProtocols::MplsInIp => "MplsInIp", // 137
+            &IpNextHeaderProtocols::Manet => "Manet", // 138
+            &IpNextHeaderProtocols::Hip => "Hip", // 139
+            &IpNextHeaderProtocols::Shim6 => "Shim6", // 140
+            &IpNextHeaderProtocols::Wesp => "Wesp", // 141
+            &IpNextHeaderProtocols::Rohc => "Rohc", // 142
+            &IpNextHeaderProtocols::Test1 => "Test1", // 253
+            &IpNextHeaderProtocols::Test2 => "Test2", // 254*/
+            &IpNextHeaderProtocols::Reserved => "Reserved", // 255
+            _                          => "unknown"
```
mrmonday: I'll trust you that this is exhaustive :wink:

mrmonday:
Looks great, thanks!

@homu r+

src/packet/ip.rs
fsck: I guess [this](http://www.iana.org/assignments/protocol-numbers/protocol-numbers.xhtml) is the official reference. I could not find any discrepancies. But if anybody has any complaints I am happy to provide an addendum. ;-)

mrmonday: I originally generated all those constants from the list, so if you have them all (unless there are new entries), that should be all of them I think!

PR #224, opened 2016-10-13
Add comment: syntex version must match pnet_macros

k4rtik:
From the discussion with @faern and levi on #libpnet

### DIFF
--- a/docs/using_packet.md
+++ b/docs/using_packet.md
@@ -29,7 +29,7 @@ version = "0.1.0"
 authors = ["My Name <my.email@mydomain.com>"]
 build = "build.rs"
 [build-dependencies]
-syntex = "0.42.*" # syntex version must match the version being used in pnet_macros/Cargo.toml
+syntex = "X" # where X is the version of syntex used in pnet_macros/Cargo.toml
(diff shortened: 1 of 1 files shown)


docs/using_packet.md
```diff
@@ -29,7 +29,7 @@ version = "0.1.0"
 authors = ["My Name <my.email@mydomain.com>"]
 build = "build.rs"
 [build-dependencies]
-syntex = "0.31.*"
+syntex = "0.42.*" # syntex version must match the version being used in pnet_macros/Cargo.toml
```
faern: Specifying a number here at all is just doomed to become outdated sooner or later. IMO it's better to say something like
`syntex = "X" # Where X must be the syntex version being used in pnet_macros/Cargo.toml`

k4rtik: That makes sense. Thanks for taking a look @faern.

mrmonday: Seems like the kind of thing that would be nice to have as part of a release script at some point in the future - we could automatically update the document with the right version.

mrmonday:
@homu r+

Thanks for the fix!

PR #225, opened 2016-10-13
Checksum flexibility

faern:
This PR more or less does three small things:

1. Make it possible to send `extra_data` to an `ipvX_checksum` function. This removes one copy when constructing packets that are fragmented. You have all the payload in a vec but the header is not attached ahead of it. With the previous checksum function one had to allocate new memory, build the header and then copy in all the payload before calculating the checksum. This fix introduces the slight annoyance of having to send `&[]` as the second argument to the checksum function whenever you want to calculate the checksum but does not have any `extra_data`. Is this OK or should I split it into two functions?

2. Removes the `next_level_protocol` argument from `udp::ipvX_checksum` and `tcp::ipv4_checksum`. Feels inappropriate to use `udp::ipv4_checksum` to calculate anything but a Udp header and similar for tcp?

3. Made `udp::ipv6_checksum` use the pretty new (and significantly faster) checksumming in `util` instead of implementing it itself.

@mrmonday

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,6 @@
 language: rust
 rust:
-    - 1.9.0
+    - 1.11.0
     - stable
     - beta
     - nightly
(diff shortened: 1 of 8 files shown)


faern:
I bumped the minimum version to 1.11.0 since that is when the `sum` function on iterators was stabilized. I don't know if you think it's worth bumping for one simple usage of that function, but I think you said before we could bump it if needed. Code can be changed if you prefer that.

mrmonday:
This looks good - is there anyway to do this with a deprecation step? As you know, I don't like breaking things straight off if we don't have to.

faern:
I rebased and forcepushed this PR. It was based on an older master commit and the history over there was getting pretty nasty (non-linear).

mrmonday:
@homu r+

Thanks :)

PR #226, opened 2016-10-13
Rename icmp_codes and icmp_types to standard

faern:
The modules `icmp_types` and `icmp_codes` does not follow the standard convention used throughout the rest of pnet. Here I converted them from snake_case to CamelCase to conform.

This is a breaking API change. However I feel this "error" can't stay forever. If you really want a period where we support the old name as well, how do you propose we do that?

@mrmonday


faern:
I solved the deprecated problem myself by adding back all the modules with the old name and the #[deprecated] flag. We can remove them after the next release, so they only exist in 0.14.0

mrmonday:
@homu r+

PR #230, opened 2016-10-24
Remove deprecated

faern:
New version released. So deprecated stuff can be removed. Nothing as uplifting as a red-only PR :wink:


faern:
@homu r+

PR #231, opened 2016-10-30
Running rustfmt on all of libpnet

faern:
To clean up all of libpnet I ran rustfmt on the entire library. A few things in src/packet are not formatted since they ended up looking very bad.


faern:
@homu r+

PR #232, opened 2016-11-22
Compiler warning cleanup

faern:
There were a lot of compiler warnings. Mostly unused imports and a few unused functions. I cleaned these up.

### DIFF
--- a/src/util.rs
+++ b/src/util.rs
@@ -10,15 +10,12 @@
 
 extern crate libc;
 
-use internal;
 use packet::PrimitiveValues;
 use packet::ip::IpNextHeaderProtocol;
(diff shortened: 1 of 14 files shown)


src/util.rs
```diff
@@ -138,68 +133,6 @@ fn str_from_mac_addr() {
                "12:34:56:78:09:ab");
 }
 
-#[cfg(target_os = "linux")]
-fn sockaddr_to_network_addr(sa: *const sockets::SockAddr) -> (Option<MacAddr>, Option<IpAddr>) {
```
faern: They have been moved to unix_interfaces.rs but for some reason they were left here as well.

PR #233, opened 2016-11-29
Add Icmp transport channel iterator

martinschleiss:
Hi,

I had the need for an ICMP packet iterator. I added it to the existing ones since I can not use
transport_channel_iterator! because of some dependencies on private modules like pnet::sockets or pnet::internal .

Best,
Martin


mrmonday:
Great, thanks!

@homu r+

PR #236, opened 2016-12-14
Allow windows msys/mingw executable variables with spaces

termquick:
Added lots of quotes around automatically found executable variables to prevent directory names being mangled in windows.


mrmonday:
Great stuff, thanks!

@homu r+

PR #237, opened 2016-12-28
Fix a typo in arp.rs

changlan:


mrmonday:
Good catch, thank you!

@homu r+

Issue #81, opened 2015-06-15
NetworkInterface addresses and masks


cortesi:
NetworkInterface would be much more useful if the interface addresses included the netmask. This might be done with a MaskedIP type that consists conceptually of a (ip, mask) tuple with some conveniences to check if a given address falls within the ip/mask combo. If you'd be willing to merge something like this, I can implement it for OSX and Linux.

mrmonday:
This would be great to have, it should be a relatively straight forwards addition - just modify the `ips` vector in `NetworkInterface` to contain tuples rather than just the IP, then modify `get_network_interfaces_impl()` to store both the `ifa_addr` and `ifa_mask` fields from the `struct ifaddrs` (more info about the struct is available from `man 3 getifaddrs`).

There shouldn't be any need for a new type for the netmask, since it's represented the same way as `IpAddr` - though perhaps newtype'ing it would be useful, I'd need to think about it some more.

It would be great if you could implement this! I look forward to your pull request.

cortesi:
I'll have a look at it this evening. I think a new type would be better than tuples - among other things, it will give us a nice way to constrain the underlying types to be (ipv4, ipv4) and (ipv6, ipv6), rather than some mixture. I personally also want a number of utilities that can sensibly be attached to the type - iteration of all addresses within a ip/netmask specification, checking whether an IP address is within an 
ip/netmask spec, and so forth.

Anyway, let me see what I come up with and we can talk about it on the pull request.

mrmonday:
You're probably right, that does sound sensible.

I'm rather busy for the next couple of days, but I'll try to review your pull as soon as I can!

mrmonday:
@cortesi Are you still interested on working on this?

Implementation has become more urgent since rust-lang/rust#27801 was deprecated and rust-lang/rust#30187 was merged.

If you're still interested I'll let you work on it, otherwise I'll implement it myself.

cortesi:
Not for the moment, I'm afraid. I decided to punt on the side-project I was
doing in Rust until the ecosystem had matured a bit more.

On Mon, 7 Dec 2015 at 01:08 Robert Clipsham <notifications@github.com>
wrote:

> @cortesi <https://github.com/cortesi> Are you still interested on working
> on this?
>
> Implementation has become more urgent since rust-lang/rust#27801
> <https://github.com/rust-lang/rust/issues/27801> was deprecated and
[...]

PR #247, opened 2017-02-23
Add support for subnet masks

petehayes102:
Addresses libpnet/libpnet#81.

I'd appreciate someone familiar with Windows to take a look at my winpcap patch. It appears that the IP Helper API doesn't identify `IpMask` as optional. Does this mean that an IP address without a subnet mask is a blank string? I have a suspicion that my implementation might need fine tuning?

### DIFF
--- a/src/datalink/mod.rs
+++ b/src/datalink/mod.rs
@@ -211,8 +211,8 @@ pub struct NetworkInterface {
     pub index: u32,
     /// A MAC address for the interface
     pub mac: Option<MacAddr>,
-    /// An IP addresses for the interface
-    pub ips: Option<Vec<IpAddr>>,
+    /// IP addresses and netmasks for the interface
+    pub ips: Option<Vec<IpNetmask>>,
     /// Operating system specific flags for the interface
     pub flags: u32,
 }
@@ -227,6 +227,15 @@ impl NetworkInterface {
     pub fn is_loopback(&self) -> bool {
         self.flags & (sockets::IFF_LOOPBACK as u32) != 0
     }
+}
+
+/// Represents an IP address and subnet mask
+#[derive(Clone, PartialEq, Eq, Debug, Hash)]
+pub struct IpNetmask {
+    /// IP address
+    pub ip: IpAddr,
+    /// Subnet mask
+    pub netmask: Option<IpAddr>,
 }
 
 /// Get a list of available network interfaces for the current machine.
--- a/src/datalink/mod.rs
+++ b/src/datalink/mod.rs
@@ -235,7 +235,7 @@ pub struct IpNetmask {
     /// IP address
     pub ip: IpAddr,
     /// Subnet mask
-    pub netmask: Option<IpAddr>,
+    pub netmask: IpAddr,
 }
 
 /// Get a list of available network interfaces for the current machine.
--- a/src/datalink/mod.rs
+++ b/src/datalink/mod.rs
@@ -212,7 +212,7 @@ pub struct NetworkInterface {
     /// A MAC address for the interface
     pub mac: Option<MacAddr>,
     /// IP addresses and netmasks for the interface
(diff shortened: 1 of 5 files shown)


src/datalink/mod.rs
```diff
@@ -211,8 +211,8 @@ pub struct NetworkInterface {
     pub index: u32,
     /// A MAC address for the interface
     pub mac: Option<MacAddr>,
-    /// An IP addresses for the interface
-    pub ips: Option<Vec<IpAddr>>,
+    /// IP addresses and netmasks for the interface
+    pub ips: Option<Vec<IpNetmask>>,
```
rushmorem: I prefer we change this to:-

```rust
pub ips: Vec<IpNetmask>,
```

mrmonday:
Looks ok, from a quick skim over - I'll take a more detailed look tomorrow morning.

src/datalink/mod.rs
petehayes102: Ok, sounds good. I'll add it to the PR.

src/datalink/unix_interfaces.rs
```diff
@@ -27,13 +27,7 @@ pub fn interfaces() -> Vec<NetworkInterface> {
...
             _ => new.mac,
         };
-        match (&mut old.ips, &new.ips) {
-            (&mut Some(ref mut old_ips), &Some(ref new_ips)) => {
-                old_ips.extend_from_slice(&new_ips[..])
-            }
-            (&mut ref mut old_ips @ None, &Some(ref new_ips)) => *old_ips = Some(new_ips.clone()),
-            _ => {}
-        };
+        old.ips.extend_from_slice(&new.ips[..]);
```
mrmonday: 🎊

src/datalink/mod.rs
```diff
@@ -229,6 +229,15 @@ impl NetworkInterface {
     }
 }
 
+/// Represents an IP address and subnet mask
+#[derive(Clone, PartialEq, Eq, Debug, Hash)]
+pub struct IpNetmask {
+    /// IP address
+    pub ip: IpAddr,
+    /// Subnet mask
+    pub netmask: Option<IpAddr>,
```
mrmonday: Is there ever an occasion where we have an IP, but don't have a netmask?

If there is "no netmask" (`None`), is that not equivalent to `Ipv4(0,0,0,0)` or `Ipv6(0,0,0,0,0,0,0,0)`?

mrmonday:
Looks good, with an answer to my small comment there.

This will probably have merge conflicts with #248, so we'll see who gets in first!

I'll make a release once they are both in.

src/datalink/mod.rs
petehayes102: Hrm. I know that when viewing output from `ifconfig` there are interfaces with no netmask, though that might just be because they have the default `0.0.0.0` et. al.? Happy to remove `Option` and set a default if it doesn't exist. I'm not sure what the underlying *Nix API assumes but the Windows API seems to think that there's always a subnet mask.

petehayes102: Just checked, and it seems that `Option` is appropriate:

> The ifa_netmask field points to a structure containing the netmask associated with ifa_addr, if applicable for the address family.  This field may contain a null pointer.

http://man7.org/linux/man-pages/man3/getifaddrs.3.html

petehayes102: Actually, it says that `ifa_addr` could also be null, so now I'm not sure what to think!

mrmonday: "if applicable for the address family" seems to be the key phrase to me - if the interface has an IPv4 or IPv6 address, a netmask will always be applicable (even if it's `0.0.0.0`/`::`).

petehayes102: No dramas. I'll remove the Option :)

mrmonday:
Great stuff, thank you!

@homu r+

PR #249, opened 2017-02-26
Add support for encoding/decoding NetworkInterface with Serde

petehayes102:
This is a bit left field, but it'd be awesome to add this for transmitting data from a remote server.


mrmonday:
Definitely a welcome addition, thanks!

@homu r+

petehayes102:
Hooray! Thanks mate :)

PR #250, opened 2017-02-27
Fuzzing

polachok:
https://github.com/rust-fuzz/cargo-fuzz just arrived and I tried to use it on libpnet

2 seconds into `cargo fuzz --fuzz-target fuzzer_script_1` and here we go:

```
INFO: Seed: 849846349
INFO: Loaded 0 modules (0 guards): 
Loading corpus dir: corpus
INFO: -max_len is not provided, using 64
#0      READ units: 2
thread '<unnamed>' panicked at 'attempt to subtract with overflow', /home/plhk/src/libpnet/fuzz/target/debug/build/pnet-470ff5253a202a9d/out/ipv4.rs:1209
note: Run with `RUST_BACKTRACE=1` for a backtrace.
==30801== ERROR: libFuzzer: deadly signal
    #0 0x7f76dc157b39  (/home/plhk/src/libpnet/fuzz/target/debug/fuzzer_script_1+0x1dbb39)
    #1 0x7f76dbfb0bc6  (/home/plhk/src/libpnet/fuzz/target/debug/fuzzer_script_1+0x34bc6)
    #2 0x7f76db43932f  (/lib/x86_64-linux-gnu/libpthread.so.0+0x1032f)
    #3 0x7f76dae84c36  (/lib/x86_64-linux-gnu/libc.so.6+0x36c36)
    #4 0x7f76dae88027  (/lib/x86_64-linux-gnu/libc.so.6+0x3a027)
    #5 0x7f76dc08c1d8  (/home/plhk/src/libpnet/fuzz/target/debug/fuzzer_script_1+0x1101d8)

NOTE: libFuzzer has rudimentary signal handlers.
      Combine libFuzzer with AddressSanitizer or similar for better crash reports.
SUMMARY: libFuzzer: deadly signal
MS: 0 ; base unit: 0000000000000000000000000000000000000000
0x0,0x0,0x0,0x0,0x0,0x0,0x0,0x0,0x0,0x0,0x0,0x0,0x0,0x0,0x0,0x0,0x0,0x0,0x0,0x0,0x0,0x0,0x0,0x0,0x0,0x0,0x0,0x0,0x0,0x0,0x0,0x0,0x0,0x0,0x0,0x0,0xa,
\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x0a
[...]

PR #251, opened 2017-03-08
fix panic on invalid payload start index

wg:
Hello, thanks for libpnet! I'm using it to parse some live network traffic and occasionally seeing a panic caused by the `payload()` function using a starting slice index beyond the bounds of the packet. Here's a fix that returns an empty slice if the starting index is invalid.


mrmonday:
Looks good, thank you for your work!

cc #250 


@homu r+

PR #255, opened 2017-04-02
Use ipnetwork crate

faern:
This converts from using the internal `IpNetmask` to the `IpNetwork` struct from the `ipnetwork` crate. The benefit of that is that we get a lot more functionality out of the struct and we don't have to maintain it ourselves. Always nice to use existing crates to increase interoperability between network crates in general.

I'm sorry I removed the serde functionality introduces in #249. Since `IpNetwork` does not support serialization it did not work. But also I did not see why we should provide this, for me, quite arbitrary functionality. Would transmitting interface metadata over the network be a common operation? They are machine specific and only valid in one point in time, so I don't see how transmitting it would be very useful. There might be use cases where one want to send out interface names or indexes or IPs. But I find it hard to see a case where one will need to send the entire struct.

For people who want to serialize this information it is still possible to create a custom, serializable, struct that copies the needed information. Guess some input from @petehayes102 is in order.

Another reason to remove serde is that pnet already take ages to build, more large dependencies are not optimal :P

### DIFF
--- a/examples/list_interfaces.rs
+++ b/examples/list_interfaces.rs
@@ -1,0 +1,27 @@
+// Copyright (c) 2014, 2015 Robert Clipsham <robert@octarineparrot.com>
+//
+// Licensed under the Apache License, Version 2.0 <LICENSE-APACHE or
+// http://www.apache.org/licenses/LICENSE-2.0> or the MIT license
+// <LICENSE-MIT or http://opensource.org/licenses/MIT>, at your
+// option. This file may not be copied, modified, or distributed
+// except according to those terms.
+
+/// This examples simply print all interfaces to stdout
+
+extern crate pnet;
+
+use pnet::datalink;
+
+fn main() {
+    for interface in datalink::interfaces() {
+        let mac = interface.mac.map(|mac| mac.to_string()).unwrap_or("N/A".to_owned());
+        println!("{}:", interface.name);
+        println!("  index: {}", interface.index);
+        println!("  flags: {}", interface.flags);
+        println!("  MAC: {}", mac);
+        println!("  IPs:");
+        for ip in interface.ips {
+            println!("    {:?}", ip);
+        }
+    }
+}
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,6 @@
 language: rust
 rust:
-    - 1.11.0
+    - 1.15.0
     - stable
     - beta
     - nightly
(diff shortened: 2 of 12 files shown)


examples/list_interfaces.rs
```diff
@@ -0,0 +1,27 @@
+// Copyright (c) 2014, 2015 Robert Clipsham <robert@octarineparrot.com>
+//
+// Licensed under the Apache License, Version 2.0 <LICENSE-APACHE or
+// http://www.apache.org/licenses/LICENSE-2.0> or the MIT license
+// <LICENSE-MIT or http://opensource.org/licenses/MIT>, at your
+// option. This file may not be copied, modified, or distributed
+// except according to those terms.
+
+/// This examples simply print all interfaces to stdout
```
mrmonday: 😍

faern: Wow.. So many typos on that row...

petehayes102:
I guess I'm a bit late to the party on this... I blame it on manflu. Anyway thanks for the heads up.

Cool that we're no longer having to maintain functionality, but it's annoying that serialization is out. I might have to PR upstream. Anyway, this is helpful when you want to gather information on a remote machine. There's lots of useful stuff in the `IpNetwork` struct and I didn't see the point in replicating it - extra overhead and such.

Also, you mentioned extra deps - `serde` was an optional dependency, meaning that if you didn't explicitly compile it in, you wouldn't get it, so compile times are not affected. Ironically your PR adds a mandatory dependency so compile times are now longer for everyone. I have no problem with this but since you brought it up I thought it was worth a mention. 😉

mrmonday:
I agree that would be nice functionality to have - perhaps upstream would accept a patch? If not, it shouldn't be too hard to implement this separately - just a bit of boilerplate, since you can't `#[derive]` it.

faern:
I have a hard time understanding that need. I don't want to hold back a legitimate need, but I think that need should be discussed before it's implemented. Here are my thoughs:

Is it a common use case to gather interface details from a remote machine and send all of it on the wire? If it's more of a very specific need, which it sounds like, then it's almost the same as saying that type `T` in crate `x` should support serialization.
People with this specialized need can easily create their own serializable structs with the actual data they need. If a protocol need a subset of the values I think it would be cleaner if modeled with those fields only, both to not bloat the protocol and to not leak unnecessary data from the remote host.

I can see how `IpNetwork` would be nice to have serializable. But `IpNetwork` is a tiny abstraction representing a very common concept, whereas `NetworkInterface` is just a collection of data valid more or less only in a pnet context.

I get you argument with compile times, fair enough. My biggest problem is maybe not the compile times, but rather the extra code.

PR #257, opened 2017-04-08
Linux: re-arm fd_sets before each pselect call

pdh11:
The pselect call modifies the fd_set passed-in, clearing bits for fds
which aren't readable. Previously, the socket fd was added only once at
construction time -- meaning that, if reading ever timed-out, the
EthernetDataLinkChannelIterator would never again report any received
packets.

Fix this by always re-initialising the fd_set before each call to
pselect. The same problem also affected the write side's calls to
pselect.

On *cursory* inspection, a similar problem may affect the BPF backend,
but I don't have a box to test that on.

### DIFF
--- a/src/datalink/linux.rs
+++ b/src/datalink/linux.rs
@@ -148,7 +148,7 @@ pub fn channel(network_interface: &NetworkInterface,
     }
 
     let fd = Arc::new(internal::FileDesc { fd: socket });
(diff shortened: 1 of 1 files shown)


mrmonday:
Looks good, thank you!

The BPF code looks like it does have the same issue, yes - would you mind applying a similar fix there if you have time?

PR #261, opened 2017-04-22
Add endianness documentation to #[packet]-generated code

### LINKED ISSUE #259: TcpPacket numerical field getters don't have the byte order specified in documentation
The documentation shows these getters having return types of `u32be` and `u16be`, but to check if a packet's destination port is not 443, `if tcp_pkt.get_destination() != 443` works while `tcp_pkt.get_destination().to_be() != 443` does not. (And yes, this is on a little endian machine :smile: )

SalmonProject:
Adds explanations of pnet's endianness handling to the uXXbe and uXXle type definitions' documentation. Adds similar explanations to the documentation that the #[packet] accessor and mutator templates generates.

Fixes #259.

### DIFF
--- a/pnet_macros/src/decorator.rs
+++ b/pnet_macros/src/decorator.rs
@@ -1195,6 +1195,32 @@ fn generate_sop_strings(operations: &[SetOperation]) -> String {
     op_strings
 }
 
+enum AccessorMutator { Accessor, Mutator }
+
+fn generate_accessor_or_mutator_comment(name: &str, ty: &str, op_type: AccessorMutator) -> String {
+    let get_or_set = match op_type { AccessorMutator::Accessor => "Get",
+                                     AccessorMutator::Mutator  => "Set" };
+    if let Some((_, endianness, end_specified)) = parse_ty(ty) {
+        if end_specified == EndiannessSpecified::Yes {
+            let return_or_want = match op_type { AccessorMutator::Accessor => "accessor returns",
+                                                 AccessorMutator::Mutator  => "mutator wants" };
+            if endianness == Endianness::Big {
+                format!("/// {get_or_set} the {name} field. This field is always stored big-endian
+                /// within the struct, but this {return_or_want} host order.",
+                get_or_set = get_or_set, name = name, return_or_want = return_or_want)
+            } else {
+                format!("/// {get_or_set} the {name} field. This field is always stored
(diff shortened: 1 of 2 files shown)


mrmonday:
Thank you for doing this! I've left a couple of comments, this looks like a good start :)

polachok:
We should probably take/return types without endianness suffix to avoid confusion, what do you think?
For example:
`fn get_checksum(&self) -> u16be` would be `fn get_checksum(&self) -> u16` and `fn set_checksum(&mut self, val: u16be)` would be `fn set_checksum(&mut self, val: u16)`

SalmonProject:
Where did you leave the comments? I can't find them.

Going in, I was considering suggesting getting rid of those suffixes - the "be" was what really caused my confusion. However, it appears that the #[packet] generation macros make heavy use of those suffixes. I wouldn't be surprised if it would take a lot of work to remove them.

mrmonday:
I really like the idea of making them return `u8`/`u16`/etc - that makes complete sense.

The suffixes only really make sense for the `#[packet]`s themselves.

You should be able to see the comments inline on the `Files Changed` tab.

SalmonProject:
Are you sure you posted those comments? I don't see anything there.

mrmonday:
Definitely there!
2017-04-25-224630_685x641_scrot

SalmonProject:
https://github.com/libpnet/libpnet/pull/261/files:

screenshot_20170425_171727

I'll respond here, and copy over if the comments become visible.

>copied code block

will do


>documentation on all types

Each of those types currently has its own page, and those pages get linked to by functions that use the type, e.g. https://octarineparrot.com/assets/libpnet/doc/pnet_macros_support/types/type.u16be.html. If someone is confused about the types that TcpPacket is using, that's where they're likely to end up first.

Of course, that's all moot if the `be` and `le` suffixes are removed from the generated code. Looking again, I guess that's not actually that hard? Just a matter of dropping the suffix when the actual function string is being written into the actual code? If that's true, I would also definitely like to see that happen.

SalmonProject:
Oh, I see why - they are "pending". I just had the same thing happen to me when commenting on changes in another project.

mrmonday:
Really sorry about the slow turnaround time on this - looks great now. Please chase me up if I've not responded in a few days.

I think I'll skip the review comments in future, I don't really see what value they add, and they clearly cause problems.

Thank you for your contributions!

PR #263, opened 2017-05-24
Correct the bit used for the 'more fragments' flag

gclinch:
RFC 791 describes the flags field as:

```
      Bit 0: reserved, must be zero
      Bit 1: (DF) 0 = May Fragment,  1 = Don't Fragment.
      Bit 2: (MF) 0 = Last Fragment, 1 = More Fragments.

          0   1   2
        +---+---+---+
        |   | D | M |
        | 0 | F | F |
        +---+---+---+
```

### DIFF
--- a/src/packet/ipv4.rs.in
+++ b/src/packet/ipv4.rs.in
@@ -22,7 +22,7 @@ pub mod Ipv4Flags {
     /// Don't Fragment flag
     pub const DontFragment: u3 = 0b010;
     /// More Fragments flag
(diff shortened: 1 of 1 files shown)


mrmonday:
Ouch!

Thanks for the fix!

@faern / @polachok, have either of you been bitten by this?  Not sure if your code deals with fragmented IP packets or not?

faern:
I do use fragmented IP packets but have not noticed this. I don't think I have used these flags for that I might have duplicated those bit fields.

PR #264, opened 2017-05-24
Update build.rs to use globbing instead of hard-coded file list.

termquick:
This would add `glob = "*"` to build-dependencies, but remove the hard-coded files list in build.rs.
Otherwise build.rs works in the same fashion as before.

The docs-example at `docs/using_packet.md` are updated to reflect the new build-dependency and its example build.rs uses a simplified version of the build.rs.

Both the example and libpnet-build.rs successfully ran the `cargo test` tests under Linux Mint 18.1 Serena / cargo-0.18.0-nightly (6f1b860 2017-02-11) for their respective source(s).

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -28,22 +28,16 @@ mod macros {
     
     pub fn expand() {
         // globbing for files to pre-process:
-        let pattern = "./**/*.rs.in";
+        let pattern = "./src/packet/**/*.rs.in";
         for entry in glob::glob( pattern ).expect("Failed to read glob pattern") {
-            match entry {
-                Ok(path) => {
-                    println!(" CMP-- {}", path.display() );
-                    //src: Path::new() = /full/path/file.rs.in
-                    let src     = Path::new( path.to_str().expect("Invalid src Specified.") );
-                    //src -> dst: Path::new() = OUT_DIR/file.rs
-                    let out_dir = env::var_os( "OUT_DIR" ).expect("Invalid OUT_DIR.");
-                    let file    = Path::new( path.file_stem().expect("Invalid file_stem.") );
-                    let dst     = Path::new( &out_dir ).join(file);
-                    let mut registry = syntex::Registry::new();
-                    pnet_macros::register(&mut registry);
-                    registry.expand("", &src, &dst).unwrap();
-                },
-                Err(_) => {},
+            if let Ok(path) = entry {
+                let src     = Path::new( path.to_str().expect("Invalid src Specified.") );
+                let out_dir = env::var_os( "OUT_DIR" ).expect("Invalid OUT_DIR.");
+                let file    = Path::new( path.file_stem().expect("Invalid file_stem.") );
+                let dst     = Path::new( &out_dir ).join(file);
+                let mut registry = syntex::Registry::new();
+                pnet_macros::register(&mut registry);
+                registry.expand("", &src, &dst).unwrap();
             }
         }
     }
--- a/build.rs
+++ b/build.rs
@@ -19,34 +19,32 @@ fn print_link_search_path() {}
 
 #[cfg(feature = "with-syntex")]
 mod macros {
-    extern crate syntex;
     extern crate pnet_macros;
-
+    extern crate syntex;
+    extern crate glob;
+    
     use std::env;
     use std::path::Path;
-
-    const FILES: &'static [&'static str] = &["ethernet.rs",
-                                             "gre.rs",
-                                             "ipv4.rs",
-                                             "ipv6.rs",
-                                             "icmp.rs",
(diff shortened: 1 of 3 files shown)


polachok:
Woah, you can't just put '*' in there! Please specify version.

termquick:
Sorry... Silly "globbing" on my part.
Restricting it to the major version branch of "0.2.*".
Thoughts?

build.rs
```diff
@@ -19,34 +19,32 @@ fn print_link_search_path() {}
...
-            let src_file = format!("src/packet/{}.in", file);
-            let src = Path::new(&src_file);
-            let dst = Path::new(&out_dir).join(file);
-
-            let mut registry = syntex::Registry::new();
-            pnet_macros::register(&mut registry);
-
-            registry.expand("", &src, &dst).unwrap();
+        // globbing for files to pre-process:
+        let pattern = "./**/*.rs.in";
```
mrmonday: I'd be happier if this only looked in `src/packet/`, as the existing script does.

build.rs
```diff
@@ -19,34 +19,32 @@ fn print_link_search_path() {}
...
+        for entry in glob::glob( pattern ).expect("Failed to read glob pattern") {
+            match entry {
+                Ok(path) => {
+                    println!(" CMP-- {}", path.display() );
```
mrmonday: No `println!`s in the build script, please

build.rs
```diff
@@ -19,34 +19,32 @@ fn print_link_search_path() {}
...
+                    //src: Path::new() = /full/path/file.rs.in
+                    let src     = Path::new( path.to_str().expect("Invalid src Specified.") );
+                    //src -> dst: Path::new() = OUT_DIR/file.rs
```
mrmonday: I don't think these commented lines add any value here.

build.rs
```diff
@@ -19,34 +19,32 @@ fn print_link_search_path() {}
...
+                    let out_dir = env::var_os( "OUT_DIR" ).expect("Invalid OUT_DIR.");
+                    let file    = Path::new( path.file_stem().expect("Invalid file_stem.") );
+                    let dst     = Path::new( &out_dir ).join(file);
+                    let mut registry = syntex::Registry::new();
+                    pnet_macros::register(&mut registry);
+                    registry.expand("", &src, &dst).unwrap();
+                },
+                Err(_) => {},
```
mrmonday: Since you're ignoring the error case, this `match` would probably be better as an `if let` instead.

docs/using_packet.md
```diff
@@ -181,27 +182,38 @@ Invoking Syntex From Your Build Script
 --------------------------------------
 
 To pull everything together, we need to invoke `syntex` within the build script.
-This is a simple build script for building just a single packet type; see the
-`build.rs` from `libpnet` for an example of how to build a bunch of them:
+This is a simple build script for recursively building files using the pnet_macros; see also the
+`build.rs` from `libpnet` as an example of how many packets are built:
```
mrmonday: I'm not sure this comment makes sense - this example previously only built a single file, it then referred the reader to the `build.rs` script for building multiple files.

I would either keep this example only using a single file, and leave the original text; remove the comment about `build.rs`; or change the comment to specify it as a more complete example - though it looks copy/pasted so maybe not?

mrmonday:
What do the build times look like before/after this? I'm weary of adding extra dependencies to the build, since we already take a long time to build.

docs/using_packet.md
termquick: Ok, I will look into either reverting the example section or clear up the grammar.
The goal of my edit was to change as little of the example as possible but still add the usage of glob.
So it would not surprise me if the grammar is wonky because of only minorly changing the description text.

termquick:
With a Linux 4-core @4GHz VM for from timing build + test times from a fresh git-clone (omitting download times), averaged over three tests:
Original ( libpnet commit https://github.com/termquick/libpnet/commit/c9ebe5e33db98c96b6e146ee4b04e297eeff3a38 )  --  54 seconds
With glob (termquick commit https://github.com/termquick/libpnet/commit/c9ebe5e33db98c96b6e146ee4b04e297eeff3a38 )  -- 55 seconds

build.rs
termquick: I am assuming that I should leave this comment here while removing all others.

docs/using_packet.md
termquick: I will re-write this section first, but revert to the original if this example usage is seen as too dense.

termquick:
thoughts?

mrmonday:
Hi, @termquick!

Really sorry this has taken so long - thanks for the reminder. This looks great now, thank you for your work.

PR #266, opened 2017-07-16
Added flags argument to recv_from and flag field to transport_channel_iterator

mikeecb:
I want to be able to set custom flags for calls to `sockets::recvfrom`. Specifically, I want to be able to call `iter.next()` (where `iter` is a [`pnet::transport::Ipv4TransportChannelIterator`](https://octarineparrot.com/assets/libpnet/doc/pnet/transport/struct.Ipv4TransportChannelIterator.html)) in a non-blocking fashion by passing a [**MSG_DONTWAIT** flag](https://linux.die.net/man/2/recv). This would prevent me from having to add any threads at the application level.

Although this is a pretty simple change, it might not be in the best place!

### DIFF
--- a/src/datalink/linux.rs
+++ b/src/datalink/linux.rs
@@ -312,7 +312,7 @@ impl<'a> EthernetDataLinkChannelIterator<'a> for DataLinkChannelIteratorImpl<'a>
         } else if ret == 0 {
             Err(io::Error::new(io::ErrorKind::TimedOut, "Timed out"))
         } else {
-            let res = internal::recv_from(self.pc.socket.fd, &mut self.pc.read_buffer, &mut caddr);
+            let res = internal::recv_from(self.pc.socket.fd, &mut self.pc.read_buffer, &mut caddr, 0);
             match res {
                 Ok(len) => Ok(EthernetPacket::new(&self.pc.read_buffer[0..len]).unwrap()),
                 Err(e) => Err(e),
--- a/src/internal/mod.rs
+++ b/src/internal/mod.rs
@@ -66,14 +66,15 @@ pub fn send_to(socket: sockets::CSocket,
 
 pub fn recv_from(socket: sockets::CSocket,
                  buffer: &mut [u8],
-                 caddr: *mut sockets::SockAddrStorage)
+                 caddr: *mut sockets::SockAddrStorage,
+                 flags: i32)
     -> io::Result<usize> {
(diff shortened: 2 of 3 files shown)


mikeecb:
Not sure why Travis failed but will look into this tomorrow

mrmonday:
I think I'd rather see this done in a similar way to how the datalink layer stuff does configuration - that's a far larger change though.

I have a couple of issues with the current implementation:
 * `set_flags()` is platform specific
 * `set_flags()` takes an integer, but there are only a limited set of valid values

I don't have a problem with exposing platform specific details or internals, since it's obviously important, but it needs to be handled tactfully.

mikeecb:
@mrmonday I don't really think the `set_flags()` method is platform specific unless we presume what the flags are. Wouldn't passing a config struct instead of an int make the method _more_ platform specific?

One idea I had is to create a `RecvConfig` trait that has a `fn to_flags(&self) -> i32` method. We could leave implementation of this up to the user or implement it ourselves for several platforms? What do you think?

mrmonday:
The flags are platform specific. Take a look at the lists of supported flags:
 * Windows: https://msdn.microsoft.com/en-us/library/windows/desktop/ms740120(v=vs.85).aspx
 * FreeBSD: http://www.unix.com/man-page/FreeBSD/2/recv/
 * Linux: https://linux.die.net/man/2/recv
 * macOS: https://developer.apple.com/legacy/library/documentation/Darwin/Reference/ManPages/man2/recvfrom.2.html

Re-thinking the Config method that the datalink modules use won't work here - with datalink, you have a choice of working directly against a platform, or working generically - this doesn't exist for the transport module.

Perhaps, given your use case, a better option might be something like:
```rust
#[cfg(platforms that support nonblocking)]
fn set_nonblocking(enabled: bool) -> Result<(), ()> {
    self.flags = self.flags | MSG_DONTWAIT;

    Ok(())
}

#[cfg(platforms that do NOT support nonblocking)]
fn set_nonblocking(enabled: bool) -> Result<(), ()> {
    Err(())
}
```

It has less flexibility, but solves your problem, and side steps the ugly issues with exposing flags directly.

mrmonday:
FYI I've fixed the build on master - if you pull in the latest changes your branch should build too.

mrmonday:
Sorry for the lack of response - I missed that you had pushed new code.

Is there any chance you could rebase this so we can get it merged? I think it looks good now.

mrmonday:
Closing due to inactivity.

PR #275, opened 2017-07-22
Assorted crashes

mrmonday:
@polachok I did a quick pass of your branch - looks like there are quite a lot of non-crash fixing changes in there.

I've pulled in (I think!) the only missing crash fix here.

Are the other changes useful for libpnet? It it worth merging them into the main repo?

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -49,7 +49,7 @@ version = "0.42.*"
 optional = true
 
 [dependencies.pcap]
-version = "0.5.*"
+version = "0.6.*"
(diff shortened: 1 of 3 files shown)


mrmonday:
Looks like the TCP fuzz test fails - I'm not familiar with the TCP code, I noticed that yours was different - should I merge that in too? I take it you fixed that case?

polachok:
I'm running `cargo fuzz tcp` for some time now, it seems to be ok with 2 patches I pushed in #277.

PR #280, opened 2017-08-02
Fix documentation for icmp packets

dlrobertson:
- Fix errors
   - Time Exceeded packet structure is not in text codeblock
   - Time Exceeded documentation labeled "destination unreachable"
 - Add packet structure from [RFC 792](https://tools.ietf.org/html/rfc792)
   - Echo Reply
   - Echo Request
   - Destination Unreachable

Overall just tried to make the icmp documentation more uniform.

### DIFF
--- a/src/packet/icmp.rs.in
+++ b/src/packet/icmp.rs.in
@@ -143,8 +143,19 @@ pub mod IcmpTypes {
 }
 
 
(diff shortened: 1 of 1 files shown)


mrmonday:
Great stuff, thanks!

PR #283, opened 2017-09-02
NdpOption members should be pub

dlrobertson:
NdpOption members should be pub


mrmonday:
Looks like the CI failures are unrelated to your change. Thanks for the fix.

This keeps happening, I wonder if it's worth writing a lint to catch this automatically.

PR #285, opened 2017-09-03
extract payload length for IPv4 and IPv6 packets

maltek:
Both IPv4 and IPv6 frames carry information how large their payloads actually are.

This commit makes use of them, instead of relying on the size of the buffer that was passed in.


mrmonday:
Nice catch!

Any chance you could put together a test case for this?

maltek:
Done!

mrmonday:
Great, thanks!

PR #288, opened 2017-10-03
Make datalink layer use &[u8] instead of packet

faern:
So we talked about this on IRC. It would be nice to split up pnet into more smaller crates, since it does many fairly unrelated things. Basically I want to be able to use the `datalink` functionality without bringing in anything else. I plan on doing that is the following steps:

1. This PR: Working towards making the `datalink` module not use the packet module by using `&[u8]` and `&mut [u8]` instead of `EthernetPacket` and `MutableEthernetPacket`. This is a breaking change. Not sure if/how we would make the transition less painful, is it worth it?
2. Break out the `datalink` module to a separate crate `pnet_datalink` or `pnet-datalink`. I personally prefer the hyphen, but this repo has traditionally used underscores so I guess we will go with that.
3. If we feel the need and if there is enough motivation, do the same for the `transport` and `packet` modules, making them separate crates.

### DIFF
--- a/src/datalink/dummy.rs
+++ b/src/datalink/dummy.rs
@@ -10,8 +10,7 @@
 //! by in memory FIFO queues. Useful for writing tests.
 
 
-use datalink::{self, DataLinkChannelIterator, DataLinkReceiver,
-               DataLinkSender, NetworkInterface};
+use datalink::{self, DataLinkReceiver, DataLinkSender, NetworkInterface};
 use std::io;
 use std::sync::mpsc::{self, Receiver, Sender};
 use std::thread;
@@ -34,10 +33,10 @@ impl Config {
     /// When using this constructor `inject_handle` and `read_handle` will return `None`.
     /// Those handles must be kept track of elsewhere.
     ///
-    /// The `DataLinkChannelIterator` created by the dummy backend will read packets from
+    /// The `DataLinkReceiver` created by the dummy backend will read packets from
     /// `receiver`. Both network errors and data can be sent on this channel.
     /// When the `receiver` channel is closed (`Sender` is dropped)
-    /// `DataLinkChannelIterator::next()` will sleep forever, simlating an idle network.
+    /// `DataLinkReceiver::next()` will sleep forever, simlating an idle network.
     ///
     /// The `DataLinkSender` created by the dummy backend will send all packets sent
     /// through `build_and_send()` and `send_to()` to the `sender` channel.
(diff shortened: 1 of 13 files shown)


faern:
Not sure how we should disconnect the `EtherType` yet that is used in `ChannelType`. I think we should make the `datalink` crate only about layer2, since that is the Datalink layer, thus discarding `ChannelType` completely. But you might have other opinions.

faern:
If you want the possibility to use layer3 in `datalink` we can replace `EtherType` with `AsRef<u16>` in that module and implement `AsRef<u16> for EtherType` so it becomes convenient to use for those who use the `packet` part of `pnet`.

src/datalink/mod.rs
```diff
@@ -144,65 +145,52 @@ pub fn channel(network_interface: &NetworkInterface, configuration: Config) -> i
...
-            #[inline]
-            fn send_to(&mut self,
-                       packet: &$packet,
-                       dst: Option<NetworkInterface>)
-                -> Option<io::Result<()>>;
-        }
-    }
+
+/// Trait to enable sending $packet packets
+pub trait EthernetDataLinkSender: Send {
```
mrmonday: Could you remove the "Ethernet" prefix from this and the other traits, please?

src/datalink/mod.rs
```diff
@@ -57,6 +56,8 @@ pub mod pcap;
 
 pub mod dummy;
 
+/// Type alias for an EtherType.
+pub type EtherType = u16;
```
mrmonday: Doesn't look like this is used anywhere?

faern: Because it has the same name, so it's not part of the diff:
```rust
pub enum ChannelType {
    /// Send and receive layer 2 packets directly, including headers
    Layer2,
    /// Send and receive "cooked" packets - send and receive network layer packets
    Layer3(EtherType),
}
```

src/datalink/dummy.rs
```diff
@@ -198,28 +178,19 @@ pub fn dummy_interface(i: u8) -> NetworkInterface {
...
     use std::sync::mpsc::{self, Receiver, Sender, TryRecvError};
     use std::thread::{sleep, spawn};
     use std::time::Duration;
 
     #[test]
-    fn send_too_small_packet_size() {
-        let (_, _, mut tx, _) = create_net();
-        // Check that it fails to send with too small packet sizes
-        assert!(tx.build_and_send(1, 0, &mut |_| {}).is_none());
-    }
```
mrmonday: Did you mean to remove this test?

faern: Yes. Now it's possible to send packets of any size. They are not longer constrained by having to be a size accepted by `EthernetPacket`

PR #290, opened 2017-10-06
Unbreak the pcap datalink

dsvensson:
There was an API change in pcap crate earlier this year, which seems to have been added in 0.6.0 affecting the sendpacket parameter.

To avoid getting this breakage again I added libpcap-dev, and enabled the pcap feature. Before doing this I had to enable sudo support for the builds to work. Are you running some special kind of payed-for Travis that doesn't need the sudo-flag?

The .travis.yml syntax for APT packages only work on Debian, but the tests don't exercise the pcap parts so the build doesn't fail if it's missing. Ideally the matrix defined in .travis.yml would take OS into account and possibly disable pcap on macOS, or perhaps install it via brew.


mrmonday:
Weird, I thought we already had the sudo flag in there. Apparently not. Perhaps we didn't need it because we started using it before the non-sudo containers appeared.

Anyway, this looks good - thank you for working on it.

PR #291, opened 2017-10-07
Datalink in own crate

faern:
Move the `datalink` module to its own crate, `pnet_datalink`. To enable this a lot of other code had to be moved around. Most stuff that was used by both `datalink` and `transport` (`internal` and `sockets`) was moved to a new `pnet_base` crate for example.

The intention here is to make it possible to depend only on `pnet_datalink` and be able to communicate on that level without bringing in any other part of pnet.


appveyor.yml
```diff
@@ -10,11 +10,11 @@ environment:
     WPD_LIB_PATH: "C:/dl/wpdpack/WpdPack/Lib/Packet.lib"
     VCVARS: "C:\\Program Files (x86)\\Microsoft Visual Studio 14.0\\VC\\bin\\vcvars32.bat"
   - TARGET: x86_64-pc-windows-msvc
-    RUST_CHANNEL: 1.19.0
+    RUST_CHANNEL: 1.20.0
     WPD_LIB_PATH: "C:/dl/wpdpack/WpdPack/Lib/x64/Packet.lib"
     VCVARS: "C:\\Program Files (x86)\\Microsoft Visual Studio 14.0\\VC\\bin\\amd64\\vcvars64.bat"
   - TARGET: i686-pc-windows-msvc
-    RUST_CHANNEL: 1.19.0
+    RUST_CHANNEL: 1.20.0
```
mrmonday: It's a shame `stable` doesn't work here - or maybe it does now?

appveyor.yml
```diff
@@ -50,3 +50,5 @@ install:
 build: false
 test_script:
     - cargo test --verbose --features appveyor
+
+skip_branch_with_pr: true
```
mrmonday: 🎊

appveyor.yml
faern: I don't think it works with the way we download Rust here. In another repo we use `rustup-init.exe` to get Rust, there `stable` works. I feel it's outside the scope of this PR to experiment with that. I can try to improve the build system later at some point.

PR #292, opened 2017-10-08
Cleanup after crate split

faern:
Found it possible to clean up a little bit.

1. Remove documentation links, so it will use docs.rs automatically.
2. Some examples could use `pnet_datalink` directly, showcasing that this is possible.
3. Some unused imports and variables.
4. "Cleaning" up Cargo.toml files. This is subjective I guess. Please provide feedback on if you like this style.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -5,7 +5,6 @@ authors = [ "Robert Clipsham <robert@octarineparrot.com>" ]
 license = "MIT/Apache-2.0"
 homepage = "https://github.com/libpnet/libpnet"
 repository = "https://github.com/libpnet/libpnet"
-documentation = "http://octarineparrot.com/assets/libpnet/doc/pnet/"
 description = "Cross-platform, low level networking using the Rust programming language."
 readme = "README.md"
 keywords = ["networking", "transport", "datalink", "packet", "protocol"]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -11,9 +11,6 @@ keywords = ["networking", "transport", "datalink", "packet", "protocol"]
 categories = ["network-programming", "parser-implementations"]
 build = "build.rs"
 
-[lib]
-name = "pnet"
-
 [features]
 default = ["with-syntex"]
 nightly = ["pnet_macros_plugin"]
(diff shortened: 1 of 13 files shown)


Cargo.toml
```diff
@@ -5,16 +5,12 @@ authors = [ "Robert Clipsham <robert@octarineparrot.com>" ]
...
 repository = "https://github.com/libpnet/libpnet"
-documentation = "http://octarineparrot.com/assets/libpnet/doc/pnet/"
 description = "Cross-platform, low level networking using the Rust programming language."
 readme = "README.md"
 keywords = ["networking", "transport", "datalink", "packet", "protocol"]
 categories = ["network-programming", "parser-implementations"]
 build = "build.rs"
 
-[lib]
-name = "pnet"
```
faern: Redundant since by default the lib name is the same as the package name.

Cargo.toml
```diff
@@ -25,51 +21,26 @@ travis = []
...
-optional = true
-version = ">=0.0"
-
-[dependencies.netmap_sys]
-features = ["netmap_with_libs"]
-optional = true
-version = ">=0.0"
-
-[dependencies.libc]
-version = "0.2.*"
```
faern: A lot of the dependencies were very simple, just a version. I do not see any reason to have them under individual blocks like this. Just makes this file super long.

pnet_macros_support/Cargo.toml
```diff
@@ -6,13 +6,6 @@ authors = ["Robert Clipsham <robert@octarineparrot.com>"]
...
-documentation = "http://octarineparrot.com/assets/libpnet/doc/pnet/"
 description = "Support library for libpnet_macros"
 readme = "../README.md"
 keywords = ["networking", "bitfields", "packet", "protocol"]
-
-[lib]
-name = "pnet_macros_support"
-
-[features]
-travis = []
```
faern: This feature was not used for anything from what I could find.

src/lib.rs
```diff
@@ -102,10 +102,6 @@
 //! }
 //! ```
 
-#![crate_name = "pnet"]
-#![crate_type = "rlib"]
-#![crate_type = "dylib"]
```
faern: Do these have a reason for being here that I missed? My guess was that they were legacy from before you used cargo?

Cargo.toml
```diff
@@ -25,51 +21,26 @@ travis = []
...
-[dependencies.pnet_macros_support]
-path = "pnet_macros_support"
-version = ">=0.1"
 
 [dev-dependencies]
 time = ">=0.1"
 
 [build-dependencies]
-glob = "0.2.*"
-syntex = { version = "0.42.*", optional = true }
```
faern: `0.x.*` is equivalent to `0.x`. I guess it's subjective which one is preferred. But I have not seen the asterisk version used almost anywhere. Not a big deal for me, it just happened to be removed when I was at it.

benches/rs_receiver.rs
```diff
@@ -7,28 +7,28 @@
 // except according to those terms.
 
 
-extern crate pnet;
+extern crate pnet_datalink;
```
faern: These benchmarks should not really be in the `benches/` directory, since they are not `#[bench]` benchmarks. This folder is special to cargo. However, it's fairly convenient to have cargo automatically compile it so it get's tested by the CI. I'll leave it in this directory for now.

Cargo.toml
mrmonday: Interesting historical point: I actually wanted `pnet` to be `libpnet`, but it was too late once it was published.

Cargo.toml
mrmonday: It is mostly this way because historically I had no idea how to work Cargo 😉 

Thanks for cleaning it up.

Cargo.toml
mrmonday: I prefer the `.*` syntax, because it's far clearer (to me at least) which versions are supported.

When I read `version = "0.42"` it's not clear to me whether that means "exactly 0.42.0" or "0.42.*" or even ">=0.42 && <= 1.0".

Given that most crates don't use the `*` I'm not too bothered if you want to leave it out.

Cargo.toml
faern: May I ask why? What is the difference between this crate and all other library crates not named `lib...`?

benches/rs_receiver.rs
mrmonday: My logic for putting them there was "these are benchmarks, and `benches/` is a benchmarks folder". It's a shame you can't just `cargo run` them like normal benchmarks though.

pnet_macros_support/Cargo.toml
mrmonday: It was probably a copy/paste thing.

src/lib.rs
mrmonday: The first line was needed pre-cargo.

The other two were to enable building as both static and dynamic libraries... I haven't attempted that in a really long time though, I suspect this isn't needed any more and can be handled automatically by Cargo. Might be worth checking?

Cargo.toml
faern: I see your point, it's a very valid one. But if you are not too bothered I'm going to leave it, so I don't have to change again. And so it looks more in line with what is common in the ecosystem. I have to go look at the definition of version specifications every time anyone uses the `^`, `~` or `*` because those are the ones I'm not used to :D

benches/rs_receiver.rs
faern: Well, they are not benchmarks really, they are normal programs. If we put them in `examples/` we could run them with `cargo run --example rs_receiver`

Cargo.toml
mrmonday: Much like `libpcap` is not just `pcap`, `libc` is not just `c`, `libpnet` is not just `pnet`. Except when it is now... 😛

Cargo.toml
mrmonday: I also hate all those sigils - I wish they'd removed them when they removed all the other sigils in Rust.

Leaving it is fine.

benches/rs_receiver.rs
mrmonday: That might be a good idea. What would you do with the Python/C benchmarks?

Issue #289, opened 2017-10-04
Compilation without pcap


angel-manuel:
I'm interested just in packet parsing. I would like to avoid including the pcap dependency.
I've tried to disable all features but the datalink module(i don't needed it either) fails to compile.
Is there any way?

mrmonday:
Breaking out the packet parsing into its own crate is something we'd like to achieve, but we're not there yet. @faern has started some of that work in #288.

The pcap dependency is optional, and isn't built by default - I don't know why it's being pulled in for you. If you can provide some more information we can try and figure out what's happening for you.

faern:
@angel-manuel That is definitely coming! I'm on the other side of the spectrum, I want to use `datalink` but without anything else. Give me a while and I will have broken out the datalink stuff at least and then maybe the rest will work better for you.
The goal is to also make the packet stuff it's own crate, but that might take a bit longer.

angel-manuel:
After the last commits, I can avoid datalink by commenting references on lib.rs and on Cargo.toml

I would be satisfied with it being a feature. No need for packet to be on a whole different crate.
It wouldn't solve @faern problem though.

faern:
Ah. Yeah, the `pnet_datalink` dependency should probably be a feature, true. Otherwise you can't get the packet without datalink of course! I was mostly happy I could now use datalink without packet :)

I'll put it behind a future (that is active by default).

faern:
@angel-manuel Check if https://github.com/libpnet/libpnet/tree/add-datalink-feature satisfies your needs.

angel-manuel:
@faern Yes, it works perfectly. Thanks a lot.

faern:
@angel-manuel However, we are likely not going to merge that. But at some point the package module will be split out into its own crate. You can even help do that if you would like.

PR #294, opened 2017-10-13
Make pnet_datalink dependency optional

faern:
The largest reason to split the crates up is to be able to just use a few of them. So it might be good to be able to depend on the top `pnet` crate without pulling in `datalink` if desired. The tests won't build without `pnet_datalink` now because tests depend on that, but the crate can build without it at least.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -12,7 +12,7 @@ categories = ["network-programming", "parser-implementations"]
 build = "build.rs"
 
 [features]
-default = ["with-syntex"]
+default = ["with-syntex", "pnet_datalink"]
 nightly = ["pnet_macros_plugin"]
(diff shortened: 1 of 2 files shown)


mrmonday:
I'm not really sure what the motivation for this is.

I would rather see the packet stuff moved to a `pnet_packet` crate instead, with the eventual state being `pnet` gets you everything, or `pnet_whatever` gives you a specific thing.

faern:
I agree that's a pretty good ideal situation. But we have no idea when we'll reach there.
This was the motivation: https://github.com/libpnet/libpnet/issues/289

faern:
Ok. This does not build at the moment. Will likely not have more time to fix it at the moment either.

PR #296, opened 2017-10-19
datalink: fix pcap

squeed:
* rename the "pcap" feature to "with-pcap" (because of dep conflict)
* pass through the with-pcap feature to pnet_datalink
* fix random build errors

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,7 +13,7 @@ env:
     - secure: kd+Q+IWrHUZK+BwwEh37IiR7B76yyvfjAB3Gx6roshKyAk2KmduoyLGv6v902gbLAtt/8JtPMBHZdKbMv7A1eKBsCOF/rMz1rHziuTbnFwfnx6UN+jalZZRIYmn20M7I1UfvhcvWXqvcrpo84NbhOYlXMmvI+X6HZ2rSIsYta6E=
     - VERBOSE: 1
   matrix:
-    - PNET_FEATURES="travis" PNET_MACROS_FEATURES="travis"
+    - PNET_FEATURES="travis pcap" PNET_MACROS_FEATURES="travis"
 matrix:
   allow_failures:
     - rust: nightly
--- a/pnet_datalink/src/pcap.rs
+++ b/pnet_datalink/src/pcap.rs
@@ -11,32 +11,41 @@ use std::path::Path;
 
 use self::pcap::{Active, Activated};
 
-use datalink::{DataLinkReceiver, DataLinkSender, NetworkInterface};
-use datalink::Channel::Ethernet;
+use {DataLinkReceiver, DataLinkSender, NetworkInterface};
+use Channel::Ethernet;
 
 /// Configuration for the pcap datalink backend
 #[derive(Clone, Copy, Debug, Eq, Hash, PartialEq)]
 pub struct Config {
-    /// The size of buffer to use when reading packets. Defaults to 4096
+    /// The size of buffer to use when reading packets. Must be at least
+    /// 65516 with pcap.
     pub read_buffer_size: usize,
 
     /// The read timeout. Defaults to None.
     pub read_timeout: Option<Duration>,
 }
 
-impl<'a> From<&'a datalink::Config> for Config {
-    fn from(config: &datalink::Config) -> Config {
-        Config{
+impl<'a> From<&'a super::Config> for Config {
+    fn from(config: &super::Config) -> Config {
+        let mut c = Config{
             read_buffer_size: config.read_buffer_size,
             read_timeout: config.read_timeout,
+        };
+        // pcap is unique in that the buffer size must be greater or equal to
+        // MAXIMUM_SNAPLEN, which is currently hard-coded to 65536
+        // So, just reset it to the default.
+        if c.read_buffer_size < 65536 {
+            c.read_buffer_size = Config::default().read_buffer_size;
         }
+        c
     }
 }
 
 impl Default for Config {
     fn default() -> Config {
         Config{
-            read_buffer_size: 4096,
+            // Just let pcap pick the default size
+            read_buffer_size: 0,
             read_timeout: None,
         }
     }
@@ -45,7 +54,7 @@ impl Default for Config {
 /// Create a datalink channel from the provided pcap device
 #[inline]
 pub fn channel(network_interface: &NetworkInterface,
-               config: Config) -> io::Result<datalink::Channel> {
+               config: Config) -> io::Result<super::Channel> {
     let cap = match pcap::Capture::from_device(&*network_interface.name) {
         Ok(cap) => cap,
         Err(e) => return Err(io::Error::new(io::ErrorKind::Other, e)),
@@ -74,7 +83,7 @@ pub fn channel(network_interface: &NetworkInterface,
 
 /// Create a datalink channel from a pcap file
 #[inline]
-pub fn from_file<P: AsRef<Path>>(path: P, config: Config) -> io::Result<datalink::Channel> {
+pub fn from_file<P: AsRef<Path>>(path: P, config: Config) -> io::Result<super::Channel> {
     let cap = match pcap::Capture::from_file(path) {
         Ok(cap) => cap,
         Err(e) => return Err(io::Error::new(io::ErrorKind::Other, e)),


squeed:
A related question: should the Travis test matrix include some different feature coverage?

Cargo.toml
```diff
@@ -19,6 +19,7 @@ netmap = ["netmap_sys"]
 appveyor = []
 travis = []
 with-syntex = ["syntex", "pnet_macros/with-syntex"]
+with-pcap = ["pnet_datalink/with-pcap"]
```
faern: Can be `pcap = ["pnet_datalink/pcap"]` if you simply use the feature created by the optional dependency.

pnet_datalink/src/pcap.rs
```diff
@@ -24,19 +24,25 @@ pub struct Config {
...
-        Config{
+impl<'a> From<&'a super::Config> for Config {
+    fn from(config: &super::Config) -> Config {
+        let mut c = Config{
             read_buffer_size: config.read_buffer_size,
             read_timeout: config.read_timeout,
+        };
+        // pcap buffer size must be at least 65616; if someone set
+        // the buffer too small, just override it with the default.
+        if c.read_buffer_size <= 65616 {
```
mrmonday: I'd rather not do this - if you specify a buffer too small, you should get an error in `datalink::channel()` instead, and the developer should explicitly set a better buffer size.

pnet_datalink/src/lib.rs
```diff
@@ -60,7 +61,8 @@ mod backend;
 pub mod netmap;
 
 #[cfg(feature = "pcap")]
-pub mod pcap;
+#[path = "pcap.rs"]
+pub mod backend;
```
mrmonday: I don't think `pcap` should ever be the default backend - the default backend should be libpnet's native backend on the platform.

There is an argument for having it as a fallback backend for platforms we don't have native support for - I don't know that that's a good idea though - what if we add a native backend in the future? That potentially breaks code.

.travis.yml
```diff
@@ -13,25 +13,25 @@ env:
...
-    - PNET_FEATURES="travis pcap nightly" PNET_MACROS_FEATURES="travis"
+    - PNET_FEATURES="travis with-syntex" PNET_MACROS_FEATURES="travis with-syntex"
+    - PNET_FEATURES="travis nightly clippy" PNET_MACROS_FEATURES="travis clippy"
+    - PNET_FEATURES="travis nightly" PNET_MACROS_FEATURES="travis"
 matrix:
   allow_failures:
     - rust: nightly
   exclude:
     - rust: 1.15.0
-      env: PNET_FEATURES="travis pcap nightly clippy" PNET_MACROS_FEATURES="travis clippy"
```
mrmonday: We intentionally added `pcap` to the build matrix so that it didn't break - is there a particular reason you're removing it?

pnet_datalink/src/pcap.rs
```diff
@@ -45,7 +51,7 @@ impl Default for Config {
 /// Create a datalink channel from the provided pcap device
 #[inline]
 pub fn channel(network_interface: &NetworkInterface,
-               config: Config) -> io::Result<datalink::Channel> {
+               config: Config) -> io::Result<super::Channel> {
```
mrmonday: I'm not sure how the builds passed with this oversight. Any ideas?

src/pnettest.rs
```diff
@@ -319,7 +319,7 @@ fn get_test_interface() -> datalink::NetworkInterface {
 }
 
 // FIXME Find a way to test this with netmap
-#[cfg(all(not(feature = "appveyor"), not(feature = "netmap")))]
+#[cfg(all(not(feature = "appveyor"), not(feature = "netmap"), not(feature = "pcap")))]
```
mrmonday: These used to work with pcap, do they not any more?

squeed:
@mrmonday When `pnet_datalink` was split out, the `pcap` feature (in `pnet`) ceased to have any effect, because it wasn't passed through to `pnet_datalink`. That's how the pcap code had compilation errors that were uncaught.

In other words, removing `pcap` from the test matrix is restoring already-existing behavior.

The backend selection logic does need to be rethought; especially because my next PR (when ebfull/pcap#91 is merged) will actually need to use the `pcap` library without changing the backend. Because the `pcap` dependency is *also* the "use-pcap-backend" feature, the meaning is overloaded.

Thoughts? One could add a `pcap-backend` feature.

src/pnettest.rs
squeed: No, now that `pcap` is a backend.

pnet_datalink/src/pcap.rs
squeed: Not a bad idea. It would also be nice to find a way to plumb through the backend-specific defaults with the abstract global defaults.

faern:
`pcap` also enables the pcap backend, it just does not make it default. So I think the naming of `pcap` vs `pcap-backend` would be a bit confusing. If we want a feature that makes the pcap backend the default I think it should be `pcap-as-default` or something along those lines.

squeed:
Unless I'm quite confused, the [backend selection logic](/libpnet/libpnet/blob/master/pnet_datalink/src/lib.rs#L153) is currently just whichever module is imported as "backend." So, I just tried to keep with that model.

faern:
The backend imported as `backend` becomes the default backend yes. The one used if you just do `pnet_datalink::channel()`. But we also have extra backends. Such as `pnet_datalink::dummy::channel()` or `pnet_datalink::netmap::channel()` if you have enabled that. So pcap could also be an extra backend that would be accessible through `pnet_datalink::pcap::channel()`

squeed:
I've edited the PR to bring the feature flag back to the original meaning - the pcap backend is optional and never default. And I've re-added the pcap flag to CI, which is green :-).

@mrmonday updated the buffer size logic.

pnet_datalink/src/pcap.rs
```diff
@@ -45,7 +47,11 @@ impl Default for Config {
...
+    if config.read_buffer_size < 65516 {
+        return Err(io::Error::new(io::ErrorKind::Other, "read_buffer_size too small"));
```
faern: If this explicit check is not in place, what happens then? If possible I usually prefer if the underlying implementation returns errors when something is wrong. Partly because if they have a similar check then this is just duplication of their code. Partly because if the underlying library changes this requirement then we would not also need a change here to be compatible with libpcap again.

faern: Iff we need the check here, then I think the magic number should be moved to a named static at the start of the file.

squeed: There is an error reported from pcap, but it's non-obvious:
> Activate: can't mmap rx ring: Invalid argument
One also can't distinguish between a "runtime" mmap failure (e.g. OOM) and a too-small buffer. The minimum is actually a kernel limitation, so you're right, we should not hard-code it here.

Probably the most user-friendly thing to do would be to separately wrap the results from `buffer_size`, so that we can return a more meaningful error.

squeed: Ah, looks like the logic is trivial; the minimum size basically just MAXIMUM_SNAPLEN. Should be easy and safe to validate here.

faern: Can you get that constant from pcap? I still feel it's sketchy to code a constant here to perform some input validation that should be the responsibility of pcap. I would prefer not checking here and maybe add a section in the docs for that buffer variable that it must be bigger than MAXIMUM_SNAPLEN or something?

With invalid config data a lot of the backends produce very strange errors. I don't really see it as our responsibility to fix that here. Then better to make a PR to those libraries and make their error messages better.

What do you think @mrmonday?

mrmonday: I agree with @faern - I would rather use the backend's errors. If there's particular setup we're meant to be doing but aren't, that's a different matter.

Of course, there are two scenarios here:

 * Using backend specific configuration
 * Using backend-agnostic configuration

In the first case, we should always pass through the exact values to the backend.

In the second case, I think it can be argued either way. Backends are free to ignore or use values that are applicable. I don't like silently changing values, but I think I'd be ok with detecting errors before we hit the backend.

squeed: I'm happy to restore the original behavior; one change that would be helpful going forward was a better way of merging the `lib::Config::Default` and individual backend defaults.

In the previous versions of this change, when I had `pcap` acting as the default backend, the default configuration was broken, because the `lib::Config::Default.read_buffer_size` will never work, with any foreseeable version of pcap (the snaplen will certainly never get _smaller_...). So it's an obvious condition that is trivially correctable.

My original addition of this fix was in the implementation of `From<super::Config>` for pcap. I think that's probably the right place if anywhere.

faern:
I'm OK with this PR now. but you'll have to rebase it first since @mrmonday just merged a pretty big change. When that is done and the tests pass we can merge it.

squeed:
@faern thanks a lot for your patience; rebased.

pnet_datalink/Cargo.toml
```diff
@@ -14,6 +14,7 @@ libc = "0.2"
 ipnetwork = "0.12"
 pnet_base = { path = "../pnet_base" }
 pnet_sys = { path = "../pnet_sys" }
+pcap = { version = "0.6", optional = true }
 
 pcap = { version = "0.6", optional = true }
```
faern: Now we ended up with two `pcap` here. It was also added here by the PR merged by @mrmonday recently. He specifically commented that he had to change those things for his PR, but also did not want to step on your toes. So he did not mean to copy your stuff, just get his stuff working :)

pnet_datalink/src/pcap.rs
```diff
@@ -11,32 +11,41 @@ use std::path::Path;
...
+        c
     }
 }
 
 impl Default for Config {
     fn default() -> Config {
         Config{
-            read_buffer_size: 4096,
+            // This is the libpcap default
+            read_buffer_size: 2097152,
```
faern: Where did you get this number from? I read here: https://docs.rs/pcap/0.7.0/pcap/struct.Capture.html#method.buffer_size And it says the default is `1000000`. But I have no idea if those are related. I just looked it up because I think it's good if we can get as many numbers as possible as variables from our dependency rather than hardcoding them.

faern:
@squeed No worries! I have patience, I like massaging code until it's nice no matter the time it takes :)
I would rather thank you for contributing!

pnet_datalink/src/pcap.rs
squeed: My source is this - https://github.com/the-tcpdump-group/libpcap/blob/master/pcap-linux.c#L3758. I wonder if libpcap supports sending a "0" buffer size for all OSs.

faern: If it does. And/or if their way of activating their default value is to send zero, then I think we should send zero.

If not. I think the `2 * 1024 * 1024` is much clearer than the computed value. It makes it easy to see that it is indeed 2 MiB.

.travis.yml
```diff
@@ -14,6 +14,7 @@ env:
     - VERBOSE: 1
   matrix:
     - PNET_FEATURES="travis" PNET_MACROS_FEATURES="travis"
+    - PNET_FEATURES="travis pcap" PNET_MACROS_FEATURES="travis"
```
faern: This effectively doubled the number of builds. Feels very overkill IMO. Could we maybe just run with pcap, and ignore without?

What do you think @mrmonday

faern:
I completely missed that this PR was updated again.
The master branch moved again, so a rebase is needed. We are planning a release as soon as this is merged.

.travis.yml
squeed: You're right; we can just build with pcap.

squeed:
Rebased and updated the travis config, PTAL.

faern:
The code is :+1: IMO. Just waiting for travis build config input from @mrmonday.

mrmonday:
I see no need to double the number of builds - unless we make the tests specifically target the pcap backend in that configuration we're just wasting CPU cycles.

If everyone's happy then I'm happy for this PR to be merged :)

faern:
I agree

faern:
Ah. I was waiting for that last travis build to finish. But all right :)

PR #297, opened 2017-10-22
Breakout crates

### LINKED ISSUE #246: Consider splitting packet parsers into a separate crate
For instance, the way `serde` does it (`serde` for core functionality, `serde_json` for serializing JSON data, etc) might not be a bad idea. A brief design:

- `pnet` and `pnet_macros` stay mostly the way they are, except with the `backend` module removed and replaced with a `Backend` trait or similar.
- `pnet_linux`, `pnet_netmap`, `pnet_bpf`, `pnet_pcap`, and `pnet_windows` crates provide backends.
- A crate called `pnet_crossplatform_defaults` or similar could implement `Backend` by proxying to the `pnet_linux`, `pnet_bpf`, and `pnet_windows` crates.

Thoughts?

mrmonday:
Break out packet and transport code into their own crates.

@faern please review/merge this if you think it's ok.

Closes #246


pnet_packet/Cargo.toml
```diff
@@ -0,0 +1,29 @@
...
+pnet_macros = { path = "../pnet_macros", optional = true, version = ">=0.9" }
+pnet_macros_support = { path = "../pnet_macros_support", version = ">=0.1" }
+pnet_macros_plugin = { path = "../pnet_macros_plugin", optional = true, version = ">=0.1" }
+
+syntex = { version = "0.42", optional = true }
+
+[build-dependencies]
+glob = "0.2"
+syntex = { version = "0.42", optional = true }
+pnet_macros = { path = "../pnet_macros", version = ">=0.13" }
```
faern: How come the `build-dependencies` and the normal `dependencies` depend on different versions of `pnet_macros`? Is it for a real reason, or just because they were added at different times?

I never specify version requirements on dependencies given by a local `path` at all. I have personally never had the need. But there might be one here that I don't know about. If not, it would be cleaner to just get rid of them. Since they are in the same repo one would always use the one that was the newest at any given checkout anyway.

pnet_packet/Cargo.toml
```diff
@@ -0,0 +1,29 @@
...
+repository = "https://github.com/libpnet/libpnet"
+description = "Cross-platform, binary packet parsing and manipulation"
+keywords = ["networking", "packet", "parsing", "raw"]
+categories = ["network-programming", "parser-implementations"]
+build = "build.rs"
+
+[features]
+default = ["with-syntex"]
+nightly = ["pnet_macros_plugin"]
+with-syntex = ["syntex", "pnet_macros/with-syntex"]
```
faern: By default we activate `with-syntex` and that in turn activates `pnet_macros/with-syntex`. But `pnet_macros` is an optional dependency and we don't activate that one by default. How does that work? If a feature of an optional dependency is activated, is the dependency also activated?

pnet_packet/Cargo.toml
```diff
@@ -0,0 +1,29 @@
...
+
+[dependencies]
+pnet_base = { path = "../pnet_base" }
+pnet_macros = { path = "../pnet_macros", optional = true, version = ">=0.9" }
+pnet_macros_support = { path = "../pnet_macros_support", version = ">=0.1" }
+pnet_macros_plugin = { path = "../pnet_macros_plugin", optional = true, version = ">=0.1" }
```
faern: I have not bothered to learn the difference between the three `pnet_macros.*` crates. How come this "top level" crate must depend on all of them? If they are support crates to `pnet_macros` would they not be pulled in by `pnet_macros` for us?

pnet_packet/src/lib.rs
```diff
@@ -212,3 +213,10 @@ pub mod arp;
 pub mod icmp;
 pub mod icmpv6;
 pub mod vlan;
+
+pub mod util;
+
+// Required to make sure that imports from pnet_macros work
+mod pnet_packet {
```
faern: Can't we modify `pnet_macros` to get rid of this requirement? It's a bit ugly.

pnet_transport/Cargo.toml
```diff
@@ -0,0 +1,20 @@
...
+
+[dependencies]
+libc = "0.2"
+#ipnetwork = "0.12"
+pnet_base = { path = "../pnet_base" }
+pnet_sys = { path = "../pnet_sys" }
+pnet_packet = { path = "../pnet_packet" }
+#
+#[target.'cfg(windows)'.dependencies]
+#winapi = "0.2"
```
faern: If these deps are actually not needed then I think we should remove them before we merge this.

pnet_packet/src/util.rs
```diff
@@ -0,0 +1,202 @@
...
+    type Output = [u8; 1];
+
+    fn octets(&self) -> Self::Output {
+        [*self]
+    }
+}
+
+/// Calculates a checksum. Used by ipv4 and icmp. The two bytes starting at `skipword * 2` will be
+/// ignored. Supposed to be the checksum field, which is regarded as zero during calculation.
+pub fn checksum(data: &[u8], skipword: usize) -> u16be {
```
faern: It would be nice if we could have the checksum functions somewhere easy to use without pulling in `pnet_packet` maybe? The only dependency they have on this crate is `IpNextHeaderProtocol`, which we could easily swap out for the corresponding primitive.

faern: I don't mean it as a the movement of this should block this PR. Just want to bring up the topic and see what you think. `pnet_base` maybe?

pnet_packet/Cargo.toml
mrmonday: * `pnet_macros` is the important one with all the logic for expanding `#[packet]`
 * `pnet_macros_plugin` has the `plugin = true` flag set, and registers a compiler plugin with a nightly rustc
 * `pnet_macros_support` provides type aliases for `u16le` etc. Any code generated by `pnet_macros` requires it.

`pnet_packet` requires either `pnet_macros` + syntex, or `pnet_macros_plugin` (which hasn't compiled in a long time because we target a rather old version of syntex... And I'm not sure syntex is even kept up to date with rustc now, but that's another issue...)

I think I should be able to get rid of the plugin/macros dependencies - they should only be build time dependencies. I'll check and see.

pnet_packet/Cargo.toml
mrmonday: My guess is it will activate `pnet_macros/with-syntex` when that optional dependency is used. I don't actually know though.

I'm thinking about scrapping the `with-syntex` stuff, and removing the `nightly` flag, since it hasn't worked for so long now, and I don't see that changing. It's probably easier than maintaining it. It could always be re-added later. What do you think?

pnet_packet/Cargo.toml
mrmonday: I think the versions might be needed when you publish on crates.io? If not I'll remove them. They're completely arbitrary and probably wrong - it's likely you need the latest versions of everything for it to work.

I have two theories for the different versions:
 * I just blindly copy/pasted from the old Cargo.toml's, and they probably just fell out of sync
 * There (was, not sure if there still is) a weird circular dependency between `pnet` and `pnet_macros` crates, where `pnet` requires `pnet_macros` to build, and `pnet_macros` requires `pnet` to run tests. I'll investigate this and clean it up if possible.

pnet_packet/src/lib.rs
mrmonday: I'll read up on the module system - I ported this hack from the previous code, which has been there since pre-1.0. The module system has changed a bit since then, and I have a far better understanding of Rust. Hopefully we can get rid of it.

pnet_packet/Cargo.toml
faern: Sounds good. Anything simplifying the build system is a blessing at this point :P

Also, focusing on supporting stable Rust is the best I think.

pnet_packet/src/util.rs
mrmonday: Perhaps, but that's for another pull request.

I think if you're writing your own packet parsing code you can probably handle writing checksum functions - I'm not sure that it's worth having *another* crate for this. Maybe it is *shruggs*.

pnet_transport/Cargo.toml
mrmonday: Ack, I thought I deleted these! Thanks.

pnet_packet/Cargo.toml
faern: Awesome if this could be simplified indeed! Also, it sounds like `pnet_macros_support` should be a dependency of `pnet_macros` so we don't need it here?

pnet_packet/Cargo.toml
faern: You can't publish `path` dependencies to crates.io no. Maybe you can if you have the version specified as well? I have no idea, if so then *mind blown*.

faern: When having multiple crates in the same repo like this and I publish them I temporarily change the dependency from being a `path` to being a published version (without a `path`) temporarily and publish, then just revert that from my working directory.

pnet_packet/src/util.rs
faern: I didn't mean creating yet another crate. I agree that would be overkill. I just meant that if we agreed that one crate we already have would be suitable then it could be there. But yes, it can stay here as well. I'll likely write my own for my packet crate anyway.

pnet_packet/Cargo.toml
mrmonday: According to http://doc.crates.io/specifying-dependencies.html#specifying-path-dependencies you can have both path and version - version is used on crates.io, and path is used locally.

pnet_packet/src/util.rs
mrmonday: I think, given that you're likely to write your own, so would anyone else. I think it should stay in pnet_packet, unless a motivating reason to do otherwise appears.

mrmonday:
Providing CI passes, I think I've covered everything we talked about, @faern.

pnet_packet/Cargo.toml
faern: Cool. If so I guess we should keep the versions, but update them to the latest ones.

.travis.yml
```diff
@@ -4,7 +4,6 @@ rust:
     - 1.15.0
     - stable
     - beta
-    - nightly
```
faern: Oh. When you said you would remove the nightly feature I thought you meant the feature only, not actually stop building on nightly. What stops us from building on nightly? We don't need to use the nighly magic compiler plugin stuff, but I still think we should build on nightly to make sure it builds. Or is there any reason we can't build on nightly?

Cargo.toml
```diff
@@ -27,20 +24,11 @@ ipnetwork = "0.12"
...
-pnet_macros = { path = "pnet_macros", optional = true, version = ">=0.9" }
-pnet_macros_support = { path = "pnet_macros_support", version = ">=0.1" }
-pnet_macros_plugin = { path = "pnet_macros_plugin", optional = true, version = ">=0.1" }
+pnet_transport = { path = "pnet_transport" }
+pnet_packet = { path = "pnet_packet" }
 
-clippy = { version = ">=0.0", optional = true }
-syntex = { version = "0.42", optional = true }
 pcap = { version = "0.6", optional = true }
 netmap_sys = { version = ">=0.0", optional = true, features = ["netmap_with_libs"] }
```
faern: I realize that `netmap_sys` should also be moved to `pnet_datalink` and not be a dependency of `pnet` directly. But we should keep the feature here: `netmap = ["pnet_datalink/netmap"]` just like the other PR fixing pcap.

pnet_packet/Cargo.toml
```diff
@@ -0,0 +1,20 @@
...
+categories = ["network-programming", "parser-implementations"]
+build = "build.rs"
+
+[dependencies]
+pnet_base = { path = "../pnet_base" }
+pnet_macros_support = { path = "../pnet_macros_support", version = ">=0.1" }
+
+[build-dependencies]
+glob = "0.2"
+syntex = { version = "0.42" }
```
faern: Can simply be `syntex = "0.42"` when only version is specified.

.travis.yml
mrmonday: I actually removed this by mistake. I considered adding it back, but I liked the idea of faster builds. I'll add it back.

Cargo.toml
mrmonday: I'll fix this.

It would be nice if cargo detected these kinds of things.

pnet_packet/Cargo.toml
mrmonday: Sure, I'll fix this.

.travis.yml
faern: We are going to get faster builds anyway I assume, Since you simplified the matrix of possible environments. Yeah, I don't really like the idea of not testing against nightly, basically everyone test against stable, beta and nightly for good reasons.

Cargo.toml
faern: Sure would be. But that would force cargo to actually inspect your code to see what you use. Which would force cargo to have a Rust source code parser. I don't think it does that. Maybe clippy can detect it?

PR #298, opened 2017-10-27
Fix pnet macros tests

mrmonday:
CI will throw errors everywhere, they're expected - unfortunately it's quite a lot of effort to capture them. The tests will show as passing, as expected.

Now they will at least run, where they didn't before.


mrmonday:
I think the next step for pnet_macros, short of scrapping it and doing it properly, would be to port it to use proceedural macros rather than syntex.

PR #299, opened 2017-11-03
[IMP] Add net device flag constants and update examples

LuoZijun:
Changes:

[pnet_sys/src/unix.rs](https://github.com/LuoZijun/libpnet/blob/patch-1/pnet_sys/src/unix.rs#L32-L68)
* Add network device flags constant for xnu and linux kernel.

[pnet_datalink/src/lib.rs](https://github.com/LuoZijun/libpnet/blob/patch-1/pnet_datalink/src/lib.rs#L219-L238)
* Add function for check device flag.

[examples/list_interfaces.rs](https://github.com/LuoZijun/libpnet/blob/patch-1/examples/list_interfaces.rs#L46)
* let `list_interfaces`  output device flag info.

[examples/packetdump.rs](https://github.com/LuoZijun/libpnet/blob/patch-1/examples/packetdump.rs#L213-L229)
* let `packetdump`  support `TUN` network device.

*Network Device Flags:*

* [XNU](https://github.com/aosm/xnu/blob/master/bsd/net/if.h#L140-L156)
* [Linux](https://github.com/torvalds/linux/blob/master/include/uapi/linux/if.h#L83-L100)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -19,7 +19,6 @@ appveyor = []
 travis = []
 
 [dependencies]
-libc = "0.2"
 ipnetwork = "0.12"
 
 pnet_base = { path = "pnet_base", version = "0.20.0" }
--- a/examples/list_interfaces.rs
+++ b/examples/list_interfaces.rs
@@ -12,14 +12,6 @@ extern crate pnet_datalink;
 
 fn main() {
     for interface in pnet_datalink::interfaces() {
-        let mac = interface.mac.map(|mac| mac.to_string()).unwrap_or("N/A".to_owned());
-        println!("{}:", interface.name);
-        println!("  index: {}", interface.index);
-        println!("  flags: {}", interface.flags);
-        println!("  MAC: {}", mac);
-        println!("  IPs:");
-        for ip in interface.ips {
-            println!("    {:?}", ip);
-        }
+        println!("{}", interface);
     }
 }
--- a/examples/packetdump.rs
+++ b/examples/packetdump.rs
@@ -15,7 +15,7 @@ use pnet::datalink::{self, NetworkInterface};
 
 use pnet::packet::Packet;
 use pnet::packet::arp::ArpPacket;
-use pnet::packet::ethernet::{EtherTypes, EthernetPacket};
+use pnet::packet::ethernet::{EtherTypes, EthernetPacket, MutableEthernetPacket};
 use pnet::packet::icmpv6::Icmpv6Packet;
 use pnet::packet::icmp::{echo_reply, echo_request, IcmpPacket, IcmpTypes};
 use pnet::packet::ip::{IpNextHeaderProtocol, IpNextHeaderProtocols};
@@ -23,6 +23,8 @@ use pnet::packet::ipv4::Ipv4Packet;
 use pnet::packet::ipv6::Ipv6Packet;
 use pnet::packet::tcp::TcpPacket;
 use pnet::packet::udp::UdpPacket;
+use pnet::util::MacAddr;
+
 use std::env;
 use std::io::{self, Write};
 use std::process;
@@ -184,18 +186,21 @@ fn handle_arp_packet(interface_name: &str, ethernet: &EthernetPacket) {
     }
 }
 
-fn handle_packet(interface_name: &str, ethernet: &EthernetPacket) {
+fn handle_ethernet_frame(interface: &NetworkInterface, ethernet: &EthernetPacket) {
+    let interface_name = &interface.name[..];
     match ethernet.get_ethertype() {
         EtherTypes::Ipv4 => handle_ipv4_packet(interface_name, ethernet),
         EtherTypes::Ipv6 => handle_ipv6_packet(interface_name, ethernet),
         EtherTypes::Arp => handle_arp_packet(interface_name, ethernet),
         _ => {
-            println!("[{}]: Unknown packet: {} > {}; ethertype: {:?} length: {}",
-                     interface_name,
-                     ethernet.get_source(),
-                     ethernet.get_destination(),
-                     ethernet.get_ethertype(),
-                     ethernet.packet().len())
+            println!(
+                "[{}]: Unknown packet: {} > {}; ethertype: {:?} length: {}",
+                interface_name,
+                ethernet.get_source(),
+                ethernet.get_destination(),
+                ethernet.get_ethertype(),
+                ethernet.packet().len()
+            )
         }
     }
 }
@@ -224,8 +229,35 @@ fn main() {
     };
 
     loop {
+        let mut buf: [u8; 1600] = [0u8; 1600];
+        let mut fake_ethernet_frame = MutableEthernetPacket::new(&mut buf[..]).unwrap();
         match rx.next() {
-            Ok(packet) => handle_packet(&interface.name[..], &EthernetPacket::new(packet).unwrap()),
+            Ok(packet) => {
+                if cfg!(target_os = "macos") 
+                    && interface.is_up()
+                    && !interface.is_broadcast() 
+                    && !interface.is_loopback()
+                    && interface.is_point_to_point() {
+                        // Maybe is TUN interface
+                        let version = Ipv4Packet::new(&packet).unwrap().get_version();
+                        if version == 4 {
+                            fake_ethernet_frame.set_destination(MacAddr(0,0,0,0,0,0));
+                            fake_ethernet_frame.set_source(MacAddr(0,0,0,0,0,0));
+                            fake_ethernet_frame.set_ethertype(EtherTypes::Ipv4);
+                            fake_ethernet_frame.set_payload(&packet);
+                            handle_ethernet_frame(&interface, &fake_ethernet_frame.to_immutable());
+                            continue;
+                        } else if version == 6 {
+                            fake_ethernet_frame.set_destination(MacAddr(0,0,0,0,0,0));
+                            fake_ethernet_frame.set_source(MacAddr(0,0,0,0,0,0));
+                            fake_ethernet_frame.set_ethertype(EtherTypes::Ipv6);
+                            fake_ethernet_frame.set_payload(&packet);
+                            handle_ethernet_frame(&interface, &fake_ethernet_frame.to_immutable());
+                            continue;
+                        }
+                }
+                handle_ethernet_frame(&interface, &EthernetPacket::new(packet).unwrap());
+            },
             Err(e) => panic!("packetdump: unable to receive packet: {}", e),
         }
     }
--- a/pnet_datalink/Cargo.toml
+++ b/pnet_datalink/Cargo.toml
@@ -10,10 +10,10 @@ keywords = ["networking", "datalink", "ethernet", "raw"]
 categories = ["network-programming"]
 
 [dependencies]
-libc = "0.2"
+libc = "0.2.34"
 ipnetwork = "0.12"
 pnet_base = { path = "../pnet_base", version = "0.20.0" }
-pnet_sys = { path = "../pnet_sys", version = "0.20.0" }
+pnet_sys = { path = "../pnet_sys", version = "0.20.1" }
 
 pcap = { version = "0.6", optional = true }
 netmap_sys = { version = ">=0.0", optional = true, features = ["netmap_with_libs"] }
--- a/pnet_datalink/src/lib.rs
+++ b/pnet_datalink/src/lib.rs
@@ -211,10 +211,69 @@ impl NetworkInterface {
     pub fn mac_address(&self) -> MacAddr {
         self.mac.unwrap()
     }
-
+    
+    pub fn is_up(&self) -> bool {
+        self.flags & (pnet_sys::IFF_UP as u32) != 0
+    }
+    pub fn is_broadcast(&self) -> bool {
+        self.flags & (pnet_sys::IFF_BROADCAST as u32) != 0
+    }
     /// Is the interface a loopback interface?
     pub fn is_loopback(&self) -> bool {
         self.flags & (pnet_sys::IFF_LOOPBACK as u32) != 0
+    }
+    pub fn is_point_to_point(&self) -> bool {
+        self.flags & (pnet_sys::IFF_POINTOPOINT as u32) != 0
+    }
+    pub fn is_multicast(&self) -> bool {
+        self.flags & (pnet_sys::IFF_MULTICAST as u32) != 0
+    }
+}
+
+
+impl ::std::fmt::Display for NetworkInterface {
+    fn fmt(&self, f: &mut ::std::fmt::Formatter) -> ::std::fmt::Result {
+        const FLAGS: [&'static str; 5] = ["UP", "BROADCAST", "LOOPBACK", "POINTOPOINT", "MULTICAST"];
+        let flags = if self.flags > 0 {
+            let rets = [
+                self.is_up(),
+                self.is_broadcast(),
+                self.is_loopback(), 
+                self.is_point_to_point(),
+                self.is_multicast()
+            ];
+            format!("{:X}<{}>", self.flags, 
+                rets.iter()
+                    .zip(FLAGS.iter()) 
+                    .filter(|&(ret, _)| ret == &true)
+                    .map(|(_, name)| name.to_string())
+                    .collect::<Vec<String>>()
+                    .join(","))
+        } else {
+            format!("{:X}", self.flags)
+        };
+        
+        let mac = self.mac.map(|mac| mac.to_string())
+                    .unwrap_or("N/A".to_owned());
+        let ips = if self.ips.len() > 0 {
+            format!("\n{}",
+                self.ips.iter()
+                    .map(|ip| {
+                        if ip.is_ipv4(){
+                            format!("       inet: {}", ip)
+                        } else {
+                            format!("      inet6: {}", ip)
+                        }
+                    })
+                    .collect::<Vec<String>>()
+                    .join("\n"))
+        } else {
+            "".to_string()
+        };
+
+    write!(f, "{}: flags={}
+      index: {}
+      ether: {}{}", self.name, flags, self.index, mac, ips)
     }
 }
 
--- a/pnet_sys/Cargo.toml
+++ b/pnet_sys/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "pnet_sys"
-version = "0.20.0"
+version = "0.20.1"
 authors = ["Robert Clipsham <robert@octarineparrot.com>", "Linus Färnstrand <faern@faern.net>"]
 license = "MIT/Apache-2.0"
 homepage = "https://github.com/libpnet/libpnet"
@@ -10,7 +10,7 @@ keywords = ["networking", "datalink", "ethernet", "raw"]
 categories = ["network-programming"]
 
 [dependencies]
-libc = "0.2"
+libc = "0.2.34"
 
 [target.'cfg(windows)'.dependencies]
 winapi = "0.2"
--- a/pnet_sys/src/unix.rs
+++ b/pnet_sys/src/unix.rs
@@ -27,8 +27,8 @@ pub mod public {
     pub const IPPROTO_IP: libc::c_int = libc::IPPROTO_IP;
     pub const IP_HDRINCL: libc::c_int = libc::IP_HDRINCL;
 
-    pub const IFF_LOOPBACK: libc::c_int = libc::IFF_LOOPBACK;
-
+    pub use libc::{IFF_UP, IFF_BROADCAST, IFF_LOOPBACK, IFF_POINTOPOINT, IFF_MULTICAST};
+    
     pub const INVALID_SOCKET: CSocket = -1;
 
 
--- a/pnet_transport/Cargo.toml
+++ b/pnet_transport/Cargo.toml
@@ -10,7 +10,7 @@ keywords = ["networking", "transport", "raw", "socket"]
 categories = ["network-programming"]
 
 [dependencies]
-libc = "0.2"
+libc = "0.2.34"
 pnet_base = { path = "../pnet_base", version = "0.20.0" }
-pnet_sys = { path = "../pnet_sys", version = "0.20.0" }
+pnet_sys = { path = "../pnet_sys", version = "0.20.1" }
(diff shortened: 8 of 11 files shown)


examples/packetdump.rs
```diff
@@ -225,7 +223,33 @@ fn main() {
...
+                    }
+                    if interface.is_broadcast(){
+                        // TAP
+                        handle_ethernet_packet(&interface.name[..], &EthernetPacket::new(packet).unwrap());
+                    }
+                    if interface.is_point_to_point(){
+                        // TUN
+                        let mut idx = 0;
+                        while packet[idx] == 0 {
+                            idx += 1;
```
faern: Hmm.. What is this? Is this code path triggered if used on a `tun` device? I never tried that.
Does every packet then start with a number of null bytes?

pnet_datalink/src/lib.rs
```diff
@@ -216,6 +216,26 @@ impl NetworkInterface {
     pub fn is_loopback(&self) -> bool {
         self.flags & (pnet_sys::IFF_LOOPBACK as u32) != 0
     }
+    #[cfg(any(target_os = "macos", target_os = "linux"))]
+    pub fn is_up(&self) -> bool {
```
faern: Are these not valid for other unix distributions? Could we not just tag them with `#[cfg(unix)]`?

pnet_sys/src/unix.rs
```diff
@@ -28,7 +28,59 @@ pub mod public {
...
+    }
+
+    #[cfg(target_os = "linux")]
+    mod device_flags {
+        use super::libc;
+        // Linux
+        // https://github.com/torvalds/linux/blob/master/include/uapi/linux/if.h#L83-L100
+        // http://man7.org/linux/man-pages/man7/netdevice.7.html
+        pub const IFF_UP: lib::c_int          = 1<<0;   //  Interface is running
+        pub const IFF_BROADCAST: lib::c_int   = 1<<1;   //  Valid broadcast address set.
```
faern: It looks like these constants are already defined in `libc`. Do we really need to redefine them?

pnet_datalink/src/lib.rs
LuoZijun: i think that will be work on other bsd-like system too, 
but i'm not sure.

pnet_sys/src/unix.rs
LuoZijun: you are right, i will remove `linux` constants.

examples/packetdump.rs
LuoZijun: 1. yes, packet from `TUN` network device.
2. for this lib, yes, because `word aligned` (https://github.com/libpnet/libpnet/blob/master/pnet_datalink/src/bpf.rs#L79) .

pnet_sys/src/unix.rs
```diff
@@ -28,7 +28,46 @@ pub mod public {
     pub const IP_HDRINCL: libc::c_int = libc::IP_HDRINCL;
 
     pub const IFF_LOOPBACK: libc::c_int = libc::IFF_LOOPBACK;
+    
+        #[cfg(target_os = "macos")]
+    mod device_flags {
```
faern: There is some indentation issue here

examples/arp_packet.rs
```diff
@@ -51,42 +58,65 @@ fn main() {
...
-        },
+        }
     };
 
     let target_mac: Result<MacAddr, ParseMacAddrErr> = match env::args().nth(4) {
         Some(n) => n.parse(),
         None => {
-            writeln!(io::stderr(), "USAGE: packetdump <NETWORK INTERFACE> <SOURCE IP> <TARGET IP> <TARGET MAC>").unwrap();
+            writeln!(
+                io::stderr(),
```
faern: There is an `eprintln!` macro stabilized now. However it does not work for the oldest Rust we say we support with libpnet. So to use that we would need to bump the minimum required version to 1.19. Not saying we need to do that. Just leaving that comment.

faern:
This does not build on Windows. Please make all CI builds pass before we can consider merging.

examples/list_interfaces.rs
```diff
@@ -10,16 +10,46 @@
 
 extern crate pnet_datalink;
 
+fn get_flags(interface: &pnet_datalink::NetworkInterface) -> String {
+    let mut flags: Vec<String> = vec![];
+    if interface.is_loopback() {
+        flags.push(String::from("LOOPBACK"));
+    }
+    if cfg!(any(target_os = "macos", target_os = "linux")) {
```
mrmonday: I'd rather not have this behind a `cfg!()` - these methods should work on all OSes.

examples/packetdump.rs
```diff
@@ -172,30 +190,43 @@ fn handle_ipv6_packet(interface_name: &str, ethernet: &EthernetPacket) {
...
         _ => {
-            println!("[{}]: Unknown packet: {} > {}; ethertype: {:?} length: {}",
-                     interface_name,
-                     ethernet.get_source(),
-                     ethernet.get_destination(),
-                     ethernet.get_ethertype(),
-                     ethernet.packet().len())
+            if cfg!(any(target_os = "macos", target_os = "linux")) {
+                let data = &ethernet.payload();
+                let version = data[0] >> 4;
```
mrmonday: There should be no manual bit shifting in code that uses `pnet_packet`.

pnet_datalink/src/lib.rs
mrmonday: I'd like these to not have any `#[cfg]` stuff. The methods are applicable on all platforms, and should be available on all platforms.

Perhaps these should return a tri-state, true/false/unknown instead?

pnet_sys/src/unix.rs
```diff
@@ -28,7 +28,46 @@ pub mod public {
...
+    
+    #[cfg(target_os = "macos")]
+    mod device_flags {
+        use super::libc;
+        // XNU
+        /// https://github.com/aosm/xnu/blob/master/bsd/net/if.h#L140-L156
+        pub const IFF_UP: libc::c_int          = 0x1;       //  interface is up
+        pub const IFF_BROADCAST: libc::c_int   = 0x2;       //  broadcast address valid
+        pub const IFF_DEBUG: libc::c_int       = 0x4;       //  turn on debugging
+        // pub const IFF_LOOPBACK: libc::c_int    = 0x8;       //  is a loopback net
```
mrmonday: Is there a reason this is commented out?

pnet_sys/src/unix.rs
```diff
@@ -28,7 +28,46 @@ pub mod public {
     pub const IP_HDRINCL: libc::c_int = libc::IP_HDRINCL;
 
     pub const IFF_LOOPBACK: libc::c_int = libc::IFF_LOOPBACK;
+    
+    #[cfg(target_os = "macos")]
+    mod device_flags {
```
mrmonday: Do these constants definitely not exist in the standard library?

pnet_sys/src/unix.rs
```diff
@@ -28,7 +28,46 @@ pub mod public {
...
+        use super::libc;
+        // XNU
+        /// https://github.com/aosm/xnu/blob/master/bsd/net/if.h#L140-L156
+        pub const IFF_UP: libc::c_int          = 0x1;       //  interface is up
+        pub const IFF_BROADCAST: libc::c_int   = 0x2;       //  broadcast address valid
+        pub const IFF_DEBUG: libc::c_int       = 0x4;       //  turn on debugging
```
faern: There are flags added here that is never unused by this PR itself. Such as this one, `IFF_DEBUG`, and `IFF_NOTRAILERS`. Why do we need to expose these flags if we don't use them. What can a user of this library do with them?

pnet_sys/src/unix.rs
faern: Yes. It's defined outside `device_flags` and set to the value exported by `libc`.

faern: But then again, it could be left out completely here, not just be commented out.

pnet_sys/src/unix.rs
mrmonday: I think this is fine - it's better to have the complete set for these things.

It's a huge hassle trying to add just one at a later point - you have to find the right header files for the right platform, then find the right constant, figure out the right way to translate it (there's usually more than one way, and which is best is never obvious), then figure out where it needs to go on the Rust side... A lot more effort than just doing it all at once. Even if some of the values turn out to be wrong, it's easier to fix them up than add them from scratch.

pnet_sys/src/unix.rs
mrmonday: I think I'd like it left in, with a comment saying where it's actually defined. Or alternatively, have it be an alias for the one in the standard lib. Or specify it here in addition to libc. I'm happy either way.

pnet_datalink/src/lib.rs
faern: If they *always* return `unknown` on other platforms I don't see the point in having them on those platforms.

That being said, I think there is great value in keeping the library cross platform! So there should be some work done trying to make them work on all platforms.

If they turn out to not be able to implement on some platforms (non unix it looks like for now) then I think they should not be conditionally compiled like this. Instead I think it's better to move them to a unix specific trait that is implemented for this struct. Then the trait is conditionally compiled. The difference is the user must explicitly import something with `unix::` in the path to use it, which will make it very clear they are not in cross-platform land any longer. That is how the standard library usually deals with this. See these exemples:
* https://doc.rust-lang.org/stable/std/os/unix/fs/trait.FileExt.html
* https://doc.rust-lang.org/stable/std/os/unix/ffi/trait.OsStrExt.html
* And some other libraries: https://docs.rs/duct/0.9.2/duct/unix/trait.HandleExt.html (full disclosure, I'm the author of that trait in duct)

mrmonday: They can be implemented at a later point - all these flags are applicable on all platforms.

If this were platform-specific functionality, then `#[cfg]` would be fine, but since it isn't, it should be available for everything, even if it's not implemented yet.

pnet_sys/src/unix.rs
```diff
@@ -28,7 +28,46 @@ pub mod public {
...
+        pub const IFF_MULTICAST: libc::c_int   = 0x8000;    //  supports multicast
+    }
+
+    #[cfg(target_os = "linux")]
+    mod device_flags {
+        use super::libc;
+        // Linux
+        // https://github.com/torvalds/linux/blob/master/include/uapi/linux/if.h#L83-L100
+        // http://man7.org/linux/man-pages/man7/netdevice.7.html
+        pub use libc::{IFF_UP, IFF_BROADCAST, IFF_DEBUG, IFF_POINTOPOINT,
```
mrmonday: Have you made a PR to `libc` for the macOS constants? Since they exist on Linux, their equivalents should proably be added for macOS too.

pnet_datalink/src/lib.rs
faern: Just afraid it will be hard to know if one can rely on them or not. Making all types `Option<bool>` becomes messy. Afraid they will be left unimplemented on some platforms forever and then we have complicated the type for no reason. Forcing Linux and macOS implementers to add extra error handling code even if it's 100% certain it will work on their platforms. Which is why I prefer the trait approach. They can still be moved to the general type as soon as they are available for all platforms, an implementer only have to remove the trait import.

pnet_sys/src/unix.rs
faern: It's defined just a few lines of code above here. And it can't be defined here as well, since all these constants are exported out and that would collide.

pnet_sys/src/unix.rs
faern: Would be awesome if we can find that. Not optimal to have to maintain them ourselves.

pnet_sys/src/unix.rs
faern: Getting it merged upstream would be gold.

pnet_sys/src/unix.rs
LuoZijun: @faern in my case, i just want to know net interface device type (tun or tap).

pnet_sys/src/unix.rs
LuoZijun: https://github.com/rust-lang/libc/pull/839

LuoZijun:
@faern @mrmonday 

all unix-like platforms:

https://github.com/rust-lang/libc/pull/839

windows platform:

https://github.com/libpnet/libpnet/pull/299/commits/3f29ca7c906c6f0c3347da1d33e0b7b8a158ab73

faern:
@LuoZijun Awesome that you try to get it merged upstream! But why do you do formatting in the PRs to libc? I'm pretty sure they won't accept that. It's unrelated to the features you want to add. Just push the extra constants and nothing else and you have a much higher chance of getting it merged I think.

The same applies to this PR. Having a formatter is great. But we need to figure out what/if what formatter settings we would like first. And we want to format in a separate formatting PR that does not also introduce a feature. Having formatting in a feature PR makes it hard to see exactly what the feature does.

LuoZijun:
@faern thank you let me know that, i will make a new `PR` .

faern:
@LuoZijun You don't need to open new PRs. You can just amend your commits and force push to your feature branch.

pnet_sys/src/unix.rs
faern: Now @LuoZijun got them merged in libc! Which is great! Only problem is that a version with them in has not been released to crates.io yet. Since libc do fairly frequent releases. Judging from the recent history it's more often than every month. I would suggest we maybe wait with merging this PR until we can use those constants in libc?

Until then there is still Windows support to implement to get rid of the conditional compilation for example.

pnet_sys/src/unix.rs
faern: Now when they exist upstream, after a release happens, then this would not need to be separate `device_flags` modules for `linux` and `macos`. The unix part of libpnet could just use `libc::IFF_*` directly.

LuoZijun:
@faern sure, it's just finshed.

of course, the `CI` build must be fail before we change `libc` version on `Cargo.toml` .

LuoZijun:
@faern ping ...

faern:
@LuoZijun Pong. Sorry for the delay. I will travel and be busy for a few days now, but I promise to take a look at this PR as soon as I have time!

pnet_transport/Cargo.toml
```diff
@@ -10,7 +10,7 @@ keywords = ["networking", "transport", "raw", "socket"]
 categories = ["network-programming"]
 
 [dependencies]
-libc = "0.2"
-pnet_base = { path = "../pnet_base", version = "0.20.0" }
-pnet_sys = { path = "../pnet_sys", version = "0.20.0" }
+libc        = ">=0.2.34"
+pnet_base   = { path = "../pnet_base", version = "0.20.0" }
+pnet_sys    = { path = "../pnet_sys", version = "0.20.1" }
```
faern: Please don't align-format like this. One space before the `=` and one after it is perfectly fine.

The problem with align formatting is that if we introduce a new dependency with a new name then all these lines must be changed, instead of just adding one single line. That results in more work just to update formatting and larger diffs than needed as well as broken `git blame`.

LuoZijun: okay

examples/packetdump.rs
```diff
@@ -172,30 +190,48 @@ fn handle_ipv6_packet(interface_name: &str, ethernet: &EthernetPacket) {
...
         EtherTypes::Arp => handle_arp_packet(interface_name, ethernet),
         _ => {
-            println!("[{}]: Unknown packet: {} > {}; ethertype: {:?} length: {}",
-                     interface_name,
-                     ethernet.get_source(),
-                     ethernet.get_destination(),
-                     ethernet.get_ethertype(),
-                     ethernet.packet().len())
+            if interface.is_up() && interface.is_point_to_point() {
+                // TUN
```
faern: What is the `EtherType` field set to in this case, where a TUN interface is used? Just want to know what's going on here.

pnet_datalink/src/lib.rs
```diff
@@ -211,11 +212,67 @@ impl NetworkInterface {
...
+            let rets = [
+                self.is_up(),
+                self.is_broadcast(),
+                self.is_loopback(), 
+                self.is_point_to_point(),
+                self.is_multicast()
+            ];
+            format!("{:X}<{}>", self.flags, 
+                rets.iter()
+                    .zip(FLAGS.iter())
```
faern: I find the design with the `FLAGS` global constant to be a bit fragile to changes. The order of the strings in that array must match the order of the function calls in the `rets` array here. And they are defined far away from each other. `FLAGS` is also not used anywhere else, so being a global constant does not serve the purpose of being available to lots of places.

Could you maybe move that global into this method instead?

Cargo.toml
```diff
@@ -19,7 +19,7 @@ appveyor = []
 travis = []
 
 [dependencies]
-libc = "0.2"
+libc = ">=0.2.34"
```
faern: You can actually remove this dependency completely. All usage of `libc` has now been moved to the sub-crates `pnet_sys` and `pnet_datalink`. Then also remember to remove the `extern crate libc;` in `src/lib.rs`.

pnet_sys/Cargo.toml
```diff
@@ -10,7 +10,7 @@ keywords = ["networking", "datalink", "ethernet", "raw"]
 categories = ["network-programming"]
 
 [dependencies]
-libc = "0.2"
+libc = ">=0.2.34"
```
faern: You could just set the version to `"0.2.34"`. If you read the docs on version specification you see that this is equal to `"^0.2.34"` [caret requirements](http://doc.crates.io/specifying-dependencies.html#caret-requirements). That would allow all 0.2.xyz versions of libc that has patch version *at least 34*.

When using inequality requirements like this I think everything that is above `0.2.34` is accepted. This means any future `0.3.0` release of libc would be pulled in automatically, while in reality it has a low chance of working. So specifying the dependency like you have done here might be risky for that reason.

pnet_transport/Cargo.toml
```diff
@@ -10,7 +10,7 @@ keywords = ["networking", "transport", "raw", "socket"]
 categories = ["network-programming"]
 
 [dependencies]
-libc = "0.2"
+libc = ">=0.2.34"
```
faern: No real need to bump this version. We only need `0.2.34` for the changes in `pnet_sys`. Nothing in `pnet_transport` explicitly requires the newer `libc`.

If it is left bumped, then please see my other comment on caret requirements.

pnet_datalink/Cargo.toml
```diff
@@ -10,7 +10,7 @@ keywords = ["networking", "datalink", "ethernet", "raw"]
 categories = ["network-programming"]
 
 [dependencies]
-libc = "0.2"
+libc = ">=0.2.34"
 ipnetwork = "0.12"
 pnet_base = { path = "../pnet_base", version = "0.20.0" }
 pnet_sys = { path = "../pnet_sys", version = "0.20.0" }
```
faern: Please bump the `pnet_sys` dependency to `0.20.1` so it depends on the version exporting the required constants.

pnet_datalink/Cargo.toml
```diff
@@ -10,7 +10,7 @@ keywords = ["networking", "datalink", "ethernet", "raw"]
 categories = ["network-programming"]
 
 [dependencies]
-libc = "0.2"
+libc = ">=0.2.34"
```
faern: See comment about `libc` in `pnet_transport/Cargo.toml`

pnet_sys/Cargo.toml
faern: I know this project has used `>=` dependencies before. But that was mostly for internal crates and used with a zero: `>=0`. We are working towards getting rid of those, which is why I don't want to introduce more of them.

examples/packetdump.rs
LuoZijun: `TUN` interface packet type is `IP_Packet(IPv4/IPv6)` , so they havn't `ether_type` .

in `libpnet` , `ether_type` filed must be `0` .

pnet_datalink/src/lib.rs
LuoZijun: ok.

examples/list_interfaces.rs
```diff
@@ -10,16 +10,11 @@
...
-        let mac = interface.mac.map(|mac| mac.to_string()).unwrap_or("N/A".to_owned());
-        println!("{}:", interface.name);
-        println!("  index: {}", interface.index);
-        println!("  flags: {}", interface.flags);
-        println!("  MAC: {}", mac);
-        println!("  IPs:");
-        for ip in interface.ips {
-            println!("    {:?}", ip);
+        if interface.flags > 0 {
+            println!("{}", interface);
```
faern: Sorry for not finding this earlier. But why only print the interface if it has flags? Would it not make sense to list all interfaces, even those without flags? That is what we do before this PR.

LuoZijun: beacuse i think that mean nothing.

your are right, for a lib, list all interfaces is make sense.

faern:
@LuoZijun One last thing before I merge this: Could you squash together the commit history a bit? There are multiple tiny commits basically toggling stuff back and fourth. This feature can be represented in well under 20 commits.

LuoZijun:
@faern now, all into one commit.

faern:
@LuoZijun Hmm, no? It's 24 commits now: https://github.com/libpnet/libpnet/pull/299/commits

I could just pick the changes and squash them myself. But I thought you wanted more than one commit. One for updating examples, one for adding constants and one for adding the methods to interfaces maybe.

If you look at the git graph (`git log --oneline --decorate --graph --all`) you'll see you just merged your changes, not squashed/rewrote them.

LuoZijun:
@faern ok.

LuoZijun:
@faern 

easy for merge:

```bash
git merge 06c11b00f42106280110e4b7a12e0dd52ee90345
git merge 86daff434b760406393ba216d00796cce03685b1
git merge dcd66c167a8f95b69e32a792463efc11675112a4
git merge acb295cc2ef538c85ccb5c2cdb3865fcd9bb360f
git merge 3a6032e831b4fd849e668bba849758bece3afc7f
git merge 84e9f83d45058e78a18bda1361f8a9a261074664
```

PR #304, opened 2017-11-16
pnet_sys: touch up sin6_scope_id use

### LINKED ISSUE #303: pnet_sys does not take sin6_scope_id into account when generating a sockaddr
Hi,

First off, this is a really cool set of libraries!  I've been using them for the last couple of days to gain some experience with Rust.  I ran into a problem when trying to send an Icmpv6Packet to fc02::1 (link local multicast to all nodes).  The error was an "address not routable" (I don't recall the exact error msg/code).

It appears that when populating the SockAddrStorage reference from a SocketAddr struct (pnet_sys::addr_to_sockaddr),  sin6_scope_id is not taken into account.  I have a fix (tested in my environment) and will send a PR for review shortly.

On a side note, have you given any thought to expanding the TransportSender API to handle ipv6 link local addresses?

Thanks!

kmroz:
Proposed to fix #303.

Tested in Linux (kernel 4.13.10) and OSX (kernel 17.2.0).

### DIFF
--- a/pnet_sys/src/lib.rs
+++ b/pnet_sys/src/lib.rs
@@ -138,6 +138,7 @@ pub fn addr_to_sockaddr(addr: SocketAddr,
                 (*storage).sin6_family = AF_INET6 as SockAddrFamily6;
                 (*storage).sin6_port = htons(addr.port());
                 (*storage).sin6_addr = inaddr;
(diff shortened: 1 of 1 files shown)


mrmonday:
Looks good, thanks!

Sorry for the lack of response.

PR #306, opened 2017-12-11
Fix pnet_datalink's netmap support

k3d3:
Since pnet_datalink was split into its own crate (or at least as far as I can tell), it's actually been impossible to enable netmap because the feature was not being passed to the child crate.

Once I did that, there were some code errors (little more than renaming datalink to super, and changing clone_from to clone_from_slice).

I also cleaned up some things cargo was complaining about (unused poll, mut eh).

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -13,7 +13,7 @@ build = "build.rs"
 
 [features]
 benchmark = []
-netmap = ["pnet_datalink/netmap_sys"]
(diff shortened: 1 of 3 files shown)


k3d3:
Might be related to #297

mrmonday:
Good catch, thank you!

Sorry it has taken so long to get to this.

PR #308, opened 2017-12-14
ipv4_payload_length() wrong use of header length

gracinet:
Hi, and thanks for this library.

I noticed a discrepancy between packet_size() and get_total_length().
Created a packet of length 84, and packet_size() returned 99.

After some digging, it turned out that ipv4_payload_length() did not
take into account that the header length is expressed in 32 bits words.

Apparently also (requires careful verification), ipv4_packet_test()
had a wrong assertion, too. Took also the opportunity to add a
direct assertion to compare packet_size() and
get_total_length() if I'm right to think they should always be equal.

### DIFF
--- a/pnet_packet/src/ipv4.rs.in
+++ b/pnet_packet/src/ipv4.rs.in
@@ -233,7 +233,20 @@ fn ipv4_options_length_test() {
 }
 
 fn ipv4_payload_length(ipv4: &Ipv4Packet) -> usize {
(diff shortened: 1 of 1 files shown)


mrmonday:
Thanks for the PR.

@faern would you mind taking a look at this, when you have time? You're more familiar with IP than I am.

kishiguro:
Looks good to me.  Let me merge the PR.

PR #310, opened 2018-02-06
Updated Rust Compiler version in travis and appveyor

ishanjain28:
The Rust compiler version in travis configuration is 1.15 which causes the builds to fail. 

I updated compiler version to 1.23 and now it passes build tests.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -2,19 +2,11 @@ environment:
   RUST_TEST_THREADS: 1
   matrix:
   - TARGET: x86_64-pc-windows-msvc
-    RUST_CHANNEL: 1.15.0
+    RUST_CHANNEL: 1.23.0
     WPD_LIB_PATH: "C:/dl/wpdpack/WpdPack/Lib/x64/Packet.lib"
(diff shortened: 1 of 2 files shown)


mrmonday:
Thanks for doing this - I hadn't had chance yet.

I think we should remove this line entirely - we test stable anyway.

Could you update the appveyor.yml too please? (remove 15, update 20 to latest stable)

appveyor.yml
```diff
@@ -2,19 +2,11 @@ environment:
   RUST_TEST_THREADS: 1
   matrix:
   - TARGET: x86_64-pc-windows-msvc
-    RUST_CHANNEL: 1.15.0
+    RUST_CHANNEL: stable
```
mrmonday: Unfortunately you need to use a specific version number here - the script pulls from `https://static.rust-lang.org/dist/rust-${env:RUST_CHANNEL}-${env:TARGET}.exe`, and there's no stable file in there.

ishanjain28: Oh, I am sorry. I'll fix it.

mrmonday:
Perfect, thank you!

PR #316, opened 2018-03-22
Improve MacAddr ergonomics

vorner:
* Make it Ord, so it can be used as a part of a key in
  BTreeMap/BTreeSet, or sorted.
* Make the ParseMacAddrErr actually an error, which makes it easier to
  integrate with things like error chain or failure.
* Removed a FIXME comment suggesting to use () instead of an error:
  - It is recommended never to use () as an error, but always at least
    provide empty struct (and implement Error on that), for ergonomics,
    type separation and clearer meaning.
  - There are use cases where MAC addresses come from eg. a
    configuration file, so it makes sense to do proper error handling
    around it, not only a programmer errors as the comment suggested.

### DIFF
--- a/pnet_base/src/macaddr.rs
+++ b/pnet_base/src/macaddr.rs
@@ -6,11 +6,12 @@
 // option. This file may not be copied, modified, or distributed
 // except according to those terms.
 
+use std::error::Error;
 use std::fmt;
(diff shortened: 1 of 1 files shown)


mrmonday:
Looking good, thanks for the work!

PR #320, opened 2018-03-25
make important fields in Transport{Receiver,Sender} public

the-eater:
Currently the `transport_channel_iterator!` fails outside pnet, since those fields are private


mrmonday:
Thanks for your work!

Could you please add a private `()` field to `TransportReceiver`? That will prevent creation from outside of libpnet, which gives a bit more flexibility in terms of API changes later on.

mrmonday:
Merging anyway - its been a really long time, and we're pre-1.0, so can always do breaking changes if needed.

Issue #281, opened 2017-08-13
Optional.unwrap() in winpcap causing crash


bgourlie:
https://github.com/libpnet/libpnet/blob/47f48e00282aabc005c96393930cb3408139a7f5/src/datalink/winpcap.rs#L277

I can try my hand at a fix, but I'm not sure how to handle the case where the option is Nothing.

mrmonday:
If you take a look at the preceding code, you'll notice there's an if statement to handle the case where are no packets in the buffer. Rather than attempting to handle the error case in this line, I would identify the issue with the preceding code causing the buffer to be empty, and handle the error there.

I'm happy to provide more guidance for this if you would like - I don't have a Windows box to test on right now, and it all worked last time I tried it (and, indeed, it works on CI).

bgourlie:
Yeah, interestingly, it successfully reads UDP packets for some time before crashing, so it's not as if its not working at all. I'll take a look at this later tonight and see if I can get a fix submitted.

bgourlie:
I've now seen this error out earlier in the function as well:

https://github.com/libpnet/libpnet/blob/47f48e00282aabc005c96393930cb3408139a7f5/src/datalink/winpcap.rs#L261

The error message, which is not verbatim since I forgot to copy it, was: "The operation completed successfully" and the OS return code was 0.

mrmonday:
These might help your investigation:
 * https://github.com/patmarion/winpcap/blob/master/WpcapSrc_4_1_3/Packet9x/DLL/Packet32.c#L524-L533
 * https://github.com/patmarion/winpcap/blob/master/WpcapSrc_4_1_3/Packet9x/DLL/Packet32.c#L134-L180
 * https://msdn.microsoft.com/en-us/library/windows/desktop/aa363216(v=vs.85).aspx

It's worth logging each of the parameters before calling `PacketReceivePacket()` - that way, when it fails, you can see exactly what input caused it to fail.

bgourlie:
A few questions (not apparent reading the linked docs):

- Is it really exceptional for ` winpcap::PacketReceivePacket` to return 0?
- If not, shouldn't the iterator that's being implemented return an `Option`

mrmonday:
According to the docts there:

```
If the operation fails or is pending, the return value is zero. 
```

So I believe that the error there is correct (except possibly in the pending case - that needs more investigation - might be worth opening a separate issue for that).

[...]

PR #327, opened 2018-04-19
Panic generate if packets is empty for winpcap

micmonay:
This error is not breaking the capture, I tested on Windows 10 1709


mrmonday:
Left my comments for this here: https://github.com/libpnet/libpnet/issues/281#issuecomment-382843955

Since we were already discussing it there.

mrmonday:
Looking good, thanks!

Issue #329, opened 2018-05-25
Data Link Interfaces call on Windows panics


JuxhinDB:
**libpnet version**: 0.21.0
**rustc version**: 1.26.0
**os**: Windows 10 Pro x64

---

On Windows, it appears that the following snippet always `panic!`(s)s:

```rust
extern crate pnet;
extern crate pnet_datalink;

fn send_tcp_packet() {
    let interfaces = pnet_datalink::interfaces();
}


fn main() {
    send_tcp_packet();
}
```

With the following:

```
C:\Users\JDB\Projects\learn-rust\syn-client>cargo run
   Compiling syn-client v0.1.0 (file:///C:/Users/JDB/Projects/learn-rust/syn-client)
    Finished dev [unoptimized + debuginfo] target(s) in 1.61 secs
     Running `target\debug\syn-client.exe`
thread 'main' panicked at 'FIXME [windows] unable to get interface list', C:\Users\JDB\.cargo\registry\src\github.com-1ecc6299db9ec823\pnet_datalink-0.21.0\src\winpcap.rs:335:9
note: Run with `RUST_BACKTRACE=1` for a backtrace.
error: process didn't exit successfully: `target\debug\syn-client.exe` (exit code: 101)
```

After some digging it appears to be related to [this](https://github.com/libpnet/libpnet/blob/89b7f5f916f109fe8588e02eea0bfcc6fdb84b90/pnet_datalink/src/winpcap.rs#L335) line of code.

List of local Network Interfaces:

```
Name                      InterfaceDescription                    ifIndex Status       MacAddress             LinkSpeed
----                      --------------------                    ------- ------       ----------             ---------
vEthernet (DockerNAT)     Hyper-V Virtual Ethernet Adapter #3          25 Up           00-15-5D-00-02-1A        10 Gbps
vEthernet (Default Swi... Hyper-V Virtual Ethernet Adapter             23 Up           16-15-7D-77-FF-87        10 Gbps
Bluetooth Network Con...2 Bluetooth Device (Personal Area Ne...#2      20 Disconnected 00-1A-7D-DA-71-02         3 Mbps
vEthernet (nat)           Hyper-V Virtual Ethernet Adapter #2          13 Up           00-15-5D-8A-D4-25        10 Gbps
vEthernet (ArchLinux E... Hyper-V Virtual Ethernet Adapter #4          12 Up           70-85-C2-0B-7F-DC       100 Mbps
[...]

PR #330, opened 2018-05-27
Fixed bug when pulling Windows net interfaces

JuxhinDB:
The following PR is related to https://github.com/libpnet/libpnet/issues/329

---

###### Stdout

```rust
extern crate pnet_datalink;

fn send_tcp_packet() {
    pnet_datalink::interfaces();
}


fn main() {
    send_tcp_packet();
}
```
```
$ cargo build
   Compiling pnet_datalink v0.21.0 (file:///C:/Users/JDB/Projects/libpnet/pnet_datalink)
   Compiling pnet v0.21.0 (file:///C:/Users/JDB/Projects/libpnet)
    Finished dev [unoptimized + debuginfo] target(s) in 1.71 secs
```
```
$ cargo run
    Finished dev [unoptimized + debuginfo] target(s) in 0.0 secs
     Running `target\debug\pnet.exe
```

---

###### Description

When calling `winpcap::PacketGetAdapterNames` the initial buffer length was being set to `4096`, which may not always be sufficient (i.e. Windows with multiple Hyper-V adapters), however when the call fails, the `buflen` is mutated to the correct value. 

In that case, `buf` is adjusted to me `4096` bytes + `buflen` which should always align correctly. Finally, a second call to `winpcap::PacketGetAdapterNames` is made with the _new_ correct buffer size.

--- 

PS: I'm not too proficient in `Rust` unfortunately, so the code may not turn out too nice. I tried wrapping the `winpcap` function call in a closure, however it kept on causing all sorts of reference issues that I could not solve. :-)

### DIFF
--- a/pnet_datalink/src/winpcap.rs
+++ b/pnet_datalink/src/winpcap.rs
@@ -324,16 +324,16 @@ pub fn interfaces() -> Vec<NetworkInterface> {
         }
     }
 
-    let mut buf = [0u8; 4096];
+    const IFACE_BUF_LEN: usize = 4096;
+    let mut buf = vec![0u8; IFACE_BUF_LEN];
     let mut buflen = buf.len() as u32;
 
-    // Gets list of supported adapters in form:
-    // adapter1\0adapter2\0\0desc1\0desc2\0\0
     if unsafe { winpcap::PacketGetAdapterNames(buf.as_mut_ptr() as *mut i8, &mut buflen) } == 0 {
-        // FIXME [windows] Should allocate a buffer big enough and try again
-        //        - size should be buf.len() + buflen (buflen is overwritten)
-        panic!("FIXME [windows] unable to get interface list");
+
+        buf = vec![0u8; IFACE_BUF_LEN + buflen as usize];
+        unsafe { winpcap::PacketGetAdapterNames(buf.as_mut_ptr() as *mut i8, &mut buflen) };
     }
+
 
     let buf_str = unsafe { from_utf8_unchecked(&buf) };
     let iface_names = buf_str.split("\0\0").next();


JuxhinDB:
@mrmonday -- The above PR solves the `panic!` issue, however there seems to be some other odd behavior I can't quite figure out. Take the following snippet:

```rust
    let mut buf = [0u8; 524288]; 
    let mut buflen = buf.len() as u32; 

    println!("Initial buffer size: {:?}", buflen);

    let outcome = unsafe {
        winpcap::PacketGetAdapterNames(buf.as_mut_ptr() as *mut i8, &mut buflen)
    };

    println!("PacketGetAdapterNames output: {:?}", outcome);
    println!("Required buffer size: {:?}", buflen);
    println!("Output buffer: {:?}", &buf[..]);
```

I'm creating an obnoxiously large buffer assuming that it will more than fit the network adapters, and even if it doesn't I was expecting the `winpcap::PacketGetAdapterNames` call to set `buflen` to the buffer size required anyway. Unfortunately the output does not seem to populate the buffer and not only that, `buflen` is getting set to `0` rather than the requested size.

```
Initial buffer size: 524288
PacketGetAdapterNames output: 0
Required buffer size: 0
Output buffer: [0, 0, ... , 0, 0]
```
Which seems to be a dead-end for me. Do you know if there is any available source of documentation for winpcap `PacketGetAdapterNames`?

pnet_datalink/src/winpcap.rs
```diff
@@ -324,17 +324,17 @@ pub fn interfaces() -> Vec<NetworkInterface> {
         }
     }
 
-    let mut buf = [0u8; 4096];
+    const IFACE_BUF_LEN: usize = 4096;
+    let mut buf = vec![0u8; IFACE_BUF_LEN];
```
mrmonday: I wouldn't bother with the constant - it's only used here, so doesn't really add anything

pnet_datalink/src/winpcap.rs
```diff
@@ -324,17 +324,17 @@ pub fn interfaces() -> Vec<NetworkInterface> {
...
     let mut buflen = buf.len() as u32;
 
-    // Gets list of supported adapters in form:
-    // adapter1\0adapter2\0\0desc1\0desc2\0\0
     if unsafe { winpcap::PacketGetAdapterNames(buf.as_mut_ptr() as *mut i8, &mut buflen) } == 0 {
-        // FIXME [windows] Should allocate a buffer big enough and try again
-        //        - size should be buf.len() + buflen (buflen is overwritten)
-        panic!("FIXME [windows] unable to get interface list");
+
+        buf = vec![0u8; IFACE_BUF_LEN + buflen as usize];
```
mrmonday: The new buffer length should be `buflen`, not the initial length + `buflen`?

Also, you may want to use this method, rather than allocating a new vector: https://doc.rust-lang.org/std/vec/struct.Vec.html#method.resize

pnet_datalink/src/winpcap.rs
```diff
@@ -324,17 +324,17 @@ pub fn interfaces() -> Vec<NetworkInterface> {
...
+        unsafe { winpcap::PacketGetAdapterNames(buf.as_mut_ptr() as *mut i8, &mut buflen) };
```
mrmonday: Please check the return value again here - panic if it returns `0`.

mrmonday:
Thanks for working on this!

There used to be some documentation somewhere, I can't find it though - it's possibly included with the winpcap release source.

Source can be found here (might not be the exact version used): https://github.com/SoftEtherVPN/Win10Pcap/blob/57da63acb45fb9b693c7fbf766253b556cbdcef6/Packet_dll/Packet32.c#L593-L658

If you can reliably detect permissions issues, I'd love another PR for that. I say reliably, because I've tried detecting whether the running application has sufficient privileges to do things on windows before, and it has always seemed hard to get it right in all circumstances. Might not be the case here.

With your snippet - do you get the same results if you use a vector, rather than an array?

JuxhinDB:
@mrmonday -- Nothing at all, it's been a pleasure. If only you knew how much time I spent trying to figure out how to re-allocate the buffer correctly, only for it to be so much simpler. :-) 

The buffer will now increase in size and pad the extended elements with `0`s. The reason I initially set the new buffer length to `4096 + buflen` was due to an old comment on the code that was making me scratch my head:

> size should be buf.len() + buflen (buflen is overwritten)

And I thought it would be better to be 4096 bytes too long rather than too short. But I can confirm that this now works correctly just by using `buflen`. 

I added another `panic!` as requested, however I did not set the message to `FIX ME [windows] ...` because that `panic!` should not really ever happen and if it does, it means something else blew up. 

Regarding the permission issue; I haven't had the time to look at it yet. However if you would like me to, I would be more than happy to. Or maybe help fix some other Windows related FIXME(s) for you given I have the environment setup.

mrmonday:
Looking great, thanks!

It would be great if you could look into some of the other FIXMEs if you have time - happy to mentor if you need a hand.

The most likely scenario for this to happen (I would imagine) is when the interface list changes in between the two calls - that's pretty unlikely though.

For little things like that, I can recommend searching the docs I linked for keywords, or asking in the Rust IRC channels if you can't find what you need. They also like pull requests if something is missing - if you struggle to find something, the chances are someone else will too :)

JuxhinDB:
Awesome, I'll check out some of the other FIXMEs towards the weekend and will open up additional PRs. Thanks for the help! :-)

PR #333, opened 2018-06-20
Update list_interfaces.rs

DD5HT:
updated extern crate name
fixed:
error[E0463]: can't find crate for `pnet_datalink`
 --> src/main.rs:1:1
  |
1 | extern crate pnet_datalink;
  | ^^^^^^^^^^^^^^^^^^^^^^^^^^^ can't find crate

error: aborting due to previous error

### DIFF
--- a/examples/list_interfaces.rs
+++ b/examples/list_interfaces.rs
@@ -7,10 +7,10 @@
 // except according to those terms.
 
 /// This examples simply print all interfaces to stdout
-extern crate pnet_datalink;
+extern crate pnet;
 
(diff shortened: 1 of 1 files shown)


mrmonday:
Could you explain how you're trying to build this?

It works for me, without changes:

```bash
$ cargo run --example list_interfaces 2&>1 >/dev/null && echo success
success
```

If you're trying to run this in its own project, you'll need to add pnet_datalink to your `Cargo.toml` instead of `pnet`.

PR #335, opened 2018-07-09
Make non-generic ICMP packet fields public

fdns:
Currently only the `Icmp` structure have his fields public, and the specific Icmp packets are private.


mrmonday:
Looks good, thanks!

PR #347, opened 2018-08-12
add type and conversion methods

flier:


pnet_base/src/macaddr.rs
```diff
@@ -13,15 +13,82 @@ use std::str::FromStr;
 #[cfg(feature = "serde")]
 use serde::{de, Deserialize, Deserializer, Serialize, Serializer};
 
+/// The number of bytes in an ethernet (MAC) address.
+pub const ETHER_ADDR_LEN: usize = 6;
+
+/// Structure of a 48-bit Ethernet address.
+pub type EtherAddr = [u8; ETHER_ADDR_LEN];
```
mrmonday: I don't think this should be public - we don't expose anything else similar to this.

pnet_base/src/macaddr.rs
```diff
@@ -13,15 +13,82 @@ use std::str::FromStr;
...
 pub struct MacAddr(pub u8, pub u8, pub u8, pub u8, pub u8, pub u8);
 
 impl MacAddr {
     /// Construct a new MacAddr
     pub fn new(a: u8, b: u8, c: u8, d: u8, e: u8, f: u8) -> MacAddr {
         MacAddr(a, b, c, d, e, f)
     }
+
+    /// Construct a empty MacAddr
+    pub fn empty() -> MacAddr {
```
mrmonday: I don't think `empty`/`is_empty` make sense for addresses - it would be better to use `Option<MacAddr>` if you genuinely want an "empty" address.

mrmonday:
Hi,

Really sorry it has taken so long to get back to you on these changes.

I think most of it looks fine - there are a couple of things I'm not sure about that I have left comments on.

FYI The best place to chase me if I don't get back to you is on IRC - I check it more regularly than my swamped inbox!

pnet_base/src/macaddr.rs
flier: Consider the user case, an ARP probe is an ARP request constructed with an all-zero sender IP address (SPA), it's not a `Option<MacAddr>`.

I think we need another name instead of `empty`, maybe `pub fn zero() -> MacAddr` and `pub fn is_zero() -> bool` will be better?

pnet_base/src/macaddr.rs
flier: no problem

mrmonday:
Welp. Took a really long time again. Sorry!

This looks good now, thank you :)

PR #350, opened 2018-10-31
Add option to set time-to-live on the TransportSender socket.

christoph-heiss:
Hi folks!

While using this awesome crate for a project of mine, I discovered that I need to set the time-to-live for packets which get sent.

So I thought I would quickly hack together a possibility to do so.
Of course, I am open to suggestion for improvements!


mrmonday:
Sorry it took such a long time to get around to this.

Looks good, thanks :)

PR #352, opened 2018-11-22
Add convenient method for get/set socket receive timeout

landesfeind:
Hi,
in this pull request, I created convenience functions for common task of setting or getting the timeout on the socket directly (implemented in pnet_sys). This comes in handy when one does not want to explicitly create the channels using a configuration. For example, I use libpnet to create an ICMP channel and later want to change the socket timeouts.

I furthermore added few basic tests to check if the setting/getting works.

The implementation is usable for `unix.rs` only because winapi has a slightly different API and I can't test it there because I have only Linux PC. However, this code may help already to implement the convenience functions there too.

Disclaimer: I am relatively new and my Rust may not be optimal yet. If I can improve my code to support your library, please let me know! 

Thanks a lot for your great work!

### DIFF
--- a/pnet_sys/src/lib.rs
+++ b/pnet_sys/src/lib.rs
@@ -90,7 +90,7 @@ pub fn recv_from(socket: CSocket,
 /// Set the timeout for the receiving from the socket
 #[cfg(unix)]
 pub fn set_socket_receive_timeout(socket: CSocket, t: Duration)
-    -> bool {
+    -> io::Result<Duration> {
     let ts = duration_to_timespec(t);
     let r = unsafe {
         setsockopt(socket, SOL_SOCKET, SO_RCVTIMEO,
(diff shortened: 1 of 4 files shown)


mrmonday:
Sorry it has taken a *really* long time to get back to you about this.

I from a quick look through, I think this looks fine - unfortunately it breaks the build on OS X and Windows though, so I can't merge it.

You can see the errors if you click "Details" next to the Travis/AppVeyor failures on the PR.

I don't mind there being platform specific functionality, but it can't break other platforms.

landesfeind:
Thanks a lot for your feedback. I reviewed the code and it should compile now on all platforms

JuxhinDB:
I took a look at the diff and it looks good. Tests and builds are also passing. Thanks a lot for this @landesfeind 

@mrmonday - I'll go ahead and merge. Feel free to revert and re-open if you have issues with it.

PR #354, opened 2019-01-25
fix call api GetAdaptersInfo crash

siten:
x86_64-pc-windows-msvc Throw a crash


mrmonday:
Sorry for the slow response - thanks!

PR #366, opened 2019-03-06
Improved the ARP packet example

Draphar:
Let's be honest: the ARP packet example was garbage. The sole purpose of ARP is getting the MAC address of a device in the network, but the old example **required** said MAC. The new example only takes the interface and the IP address of the device you want to know the MAC of as input. The interface IP and MAC is automatically extracted. Furthermore, the example now demonstrates the process of requesting a MAC, and *also* reading the answer.
I also made some small optimizations.


mrmonday:
Neat, thanks!

PR #375, opened 2019-06-20
updates Windows instructions in readme

nikita-skobov:
if using npcap, it must be in winpcap api compatible mode.

I struggled with this for a few minutes before trying out that option. Figured it might be worth adding to the readme for others trying to use this library on windows 10


JuxhinDB:
Thank you very much for this @nikita-skobov. Merging.

Issue #344, opened 2018-07-23
ipv6 packet sender


DaryaTsikach:
Hello,
I was trying to send ipv6 packet using pnet on transport channel at layer3. It was surprisingly for me, that send_to on layer 3 is not implemented for ipv6.  (I am trying to do this on my kubuntu 18.04 laptop)
I was wondering if I am doing something wrong? And if I am, could you please clarify how pnet is supposed to work with ipv6 packets?

Best regards!

mrmonday:
It is highly platform-dependent as to whether this works. IPv6 support at layer 3 on most operating systems is either not supported, or not well supported.

It looks like it should be possible to get this working on Linux - I'd be willing to mentor you if you would like to work on it.

I would personally recommend working with the datalink layer instead, and manually constructing the Ethernet frame. It has far better platform support, and far better performance.

Sorry I took a while to respond, and sorry I can't be of more use!

WGH-:
> I would personally recommend working with the datalink layer instead, and manually constructing the Ethernet frame. It has far better platform support, and far better performance.

Doing ARP manually is not fun, though :)

WGH-:
I tested this only on Linux.

A few gotchas:
* `IP_HDRINCL` doesn't work (`setsockopt(3, SOL_IP, IP_HDRINCL, [1], 4) = -1 ENOPROTOOPT (Protocol not available)`). It seems the working way to do this is to set proto 255 (`IPPROTO_RAW`), even though some sources claim 255 has no special meaning for raw IPv6 sockets.
* Source address is not filled in automatically.

```diff
diff --git a/pnet_transport/src/lib.rs b/pnet_transport/src/lib.rs
index 1a22417..382a17a 100644
--- a/pnet_transport/src/lib.rs
+++ b/pnet_transport/src/lib.rs
@@ -29,7 +29,7 @@ use pnet_packet::udp::UdpPacket;
 use pnet_packet::icmp::IcmpPacket;
 use pnet_packet::icmpv6::Icmpv6Packet;
 use pnet_packet::tcp::TcpPacket;
-use self::TransportChannelType::{Layer3, Layer4};
+use self::TransportChannelType::{Layer3, Layer4, Layer3Ipv6};
 use self::TransportProtocol::{Ipv4, Ipv6};
 
 use std::io;
@@ -56,6 +56,8 @@ pub enum TransportChannelType {
     Layer4(TransportProtocol),
[...]

PR #377, opened 2019-07-06
WIP: Add support for Layer 3 IPv6

WGH-:
This is WIP. I'm still planning to test it on Windows and FreeBSD, but haven't done so yet. And there're a couple of TODOs in the code.

But what bothers me the most is how to name new enum value. The existing `Layer3` is IPv4, and I wasn't able to solve this without either breaking compatibility (by refactoring layer 3 related enum values into something similar to what already exists for layer 4), or introducing badly named identifiers. This WIP has the latter - the new value is `Ipv6Layer3`.

Something that I've observed that might worth mentioning in the docs:
* Unlike IPv4, source address is not filled in automatically for IPv6 layer 3 packets.
* `HDRINC` (aka Layer3) has no effect for receiving IPv6 on Linux. IPv6 header is always stripped. If this is also the case with other OS, then there is no point adding `Ipv6TransportChannelIterator`

By the way, is there a way to construct an iterator over raw packets? Like `ipv4_iterator`, but with byte slice instead of `Ipv4Packet`. With `HDRINC` not working, it seems to be crucial for IPv6. Not to mention that it's generally useful for layer 4 protocols.

See #344 
ping @JuxhinDB

### DIFF
--- a/pnet_sys/src/unix.rs
+++ b/pnet_sys/src/unix.rs
@@ -31,11 +31,14 @@ pub mod public {
     pub const SOCK_RAW: libc::c_int = libc::SOCK_RAW;
 
     pub const SOL_SOCKET: libc::c_int = libc::SOL_SOCKET;
+    pub const SOL_IP: libc::c_int = libc::SOL_IP;
+    pub const SOL_IPV6: libc::c_int = libc::SOL_IPV6;
     pub const SO_RCVTIMEO: libc::c_int = libc::SO_RCVTIMEO;
     pub const SO_SNDTIMEO: libc::c_int = libc::SO_SNDTIMEO;
 
     pub const IPPROTO_IP: libc::c_int = libc::IPPROTO_IP;
     pub const IP_HDRINCL: libc::c_int = libc::IP_HDRINCL;
+    pub const IPV6_HDRINCL: libc::c_int = 36; // FIXME
     pub const IP_TTL: libc::c_int = libc::IP_TTL;
 
     pub use libc::{IFF_UP, IFF_BROADCAST, IFF_LOOPBACK, IFF_POINTOPOINT, IFF_MULTICAST};
--- a/pnet_sys/src/windows.rs
+++ b/pnet_sys/src/windows.rs
@@ -31,6 +31,7 @@ pub mod public {
 
     pub const IPPROTO_IP: libc::c_int = winapi::IPPROTO_IP;
     pub const IP_HDRINCL: libc::c_int = winapi::IP_HDRINCL;
+    pub const IPV6_HDRINCL: libc::c_int = winapi::IPV6_HDRINCL;
     pub const IP_TTL: libc::c_int = winapi::IP_TTL;
 
     pub const IFF_UP: libc::c_int = 0x00000001;
--- a/pnet_transport/src/lib.rs
+++ b/pnet_transport/src/lib.rs
@@ -24,12 +24,13 @@ extern crate pnet_packet;
 
 use pnet_packet::Packet;
 use pnet_packet::ip::IpNextHeaderProtocol;
+use pnet_packet::ipv6::Ipv6Packet;
 use pnet_packet::ipv4::Ipv4Packet;
 use pnet_packet::udp::UdpPacket;
 use pnet_packet::icmp::IcmpPacket;
 use pnet_packet::icmpv6::Icmpv6Packet;
 use pnet_packet::tcp::TcpPacket;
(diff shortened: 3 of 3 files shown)


JuxhinDB:
Thanks for this @WGH- -- I'll be going over the PR in more detail shortly and will provide my feedback

WGH-:
It seems that IPv6 layer 3 raw sockets on FreeBSD are not supported:
https://www.freebsd.org/cgi/man.cgi?query=ip6&apropos=0&sektion=4&manpath=FreeBSD+12.0-RELEASE&arch=default&format=html
>  When data received by the kernel are passed to the application, this header is not included in buffer, even when raw sockets are being used.  Likewise, when data are sent to the kernel for transmit from the application, the buffer is not examined for an IPv6 header: the kernel always constructs the header. To directly access IPv6 headers from received packets and specify them as part of the buffer	passed to the kernel, link-level access (bpf(4), for example) must instead be utilized.

WGH-:
As of Windows, its behaviour appears to be the same as with Linux: https://docs.microsoft.com/en-us/windows/win32/winsock/tcp-ip-raw-sockets-2

> When sending IPv6 data, an application has a choice on whether to specify the IPv6 header at the front of the outgoing datagram for the packet. <...> For IPv6, there should be no need to include the IPv6 header. If information is available using socket functions, then the IPv6 header should not be included to avoid compatibility problems in the future. These issues are discussed in RFC 3542 published by the IETF. Using the IPV6_HDRINCL socket option is not recommended and may be deprecated in future.

> For IPv6 (address family of AF_INET6), an application receives everything after the last IPv6 header in each received datagram regardless of the IPV6_HDRINCL socket option. The application does not receive any IPv6 headers using a raw socket.

WGH-:
https://www.ietf.org/rfc/rfc3542.txt

   >    All fields except the flow label in the IPv6 header that an    application might want to change (i.e., everything other than the    version number) can be modified using ancillary data and/or socket    options by the application for output.  All fields except the flow    label in a received IPv6 header (other than the version number and    Next Header fields) and all extension headers that an application    might want to know are also made available to the application as    ancillary data on input.  Hence there is no need for a socket option    similar to the IPv4 IP_HDRINCL socket option and on receipt the    application will only receive the payload i.e., the data after the    IPv6 header and all the extension headers. 

That's probably the reason why IPv6 layer 3 raw socket support is hit and miss across different OS

With that said, maybe it would make more sense to support the API described in RFC instead. It seems to be supported in Linux, Windows and FreeBSD (and likely OS X).

JuxhinDB:
@WGH- Agreed regarding working with the RFC. I'll be honest--at this point in time I have little knowledge, however I'll go through the RFC and see what I can gather. Will keep you updated.

mrmonday:
I'm afraid I'm not going to have time to review/assist in the near future - sorry.

The reason IPv6 L3/4 support is missing from libpnet is, as you're discovering, support for it varies from non-existent to kinda-sorta-works to completely broken between OSes. It's even more of a mess than the IPv4 L3/4 stuff, and that's far from pretty.

I've been tempted a few times to just kill off `pnet::transport`, I think it would be better served by a userspace IP stack based on `pnet::datalink` instead. The platform support would be better that way, and far more consistent.

I don't remember coming across that RFC back when I implemented `pnet::transport` (though it was quite a few years ago now...), so perhaps you'll have more luck than I did - good luck!

JuxhinDB:
Hey @mrmonday -- That's fine, I just requested review to maybe get a second pair of eyes on it. I'm still doing some reading on the matter and will see where it takes us.

WGH-:
> The reason IPv6 L3/4 support is missing from libpnet is,

IPv6 L4 is _not_ missing and seem to work fine, though.

JuxhinDB:
@WGH- I haven't forgotten about this PR. Unfortunately I'm just not available very much right now. Regarding naming, let's proceed with the naming scheme you have chosen so far. In the future we can align this better if we plan to introduce breaking changes.

> By the way, is there a way to construct an iterator over raw packets? Like ipv4_iterator, but with byte slice instead of Ipv4Packet. With HDRINC not working, it seems to be crucial for IPv6. Not to mention that it's generally useful for layer 4 protocols.

I don't know of anything particular yet but we could implement it. 

> With that said, maybe it would make more sense to support the API described in RFC instead. It seems to be supported in Linux, Windows and FreeBSD (and likely OS X).

I'm inclined to agree. I have started going through it, however it will be a number of weeks before I can look into it into more detail. Feel free to kickstart this.

mrmonday:
Closing due to inactivity.

PR #388, opened 2019-10-23
Fix some clippy and rustc warnings

dario23:


JuxhinDB:
Thanks for this, merged.

PR #389, opened 2019-10-25
Properly integrate macro variables in doc strings

dario23:
See
https://stackoverflow.com/questions/43353077/how-to-embed-a-rust-macro-variable-into-documentation
for a question-ified description of the problem and the solution.


JuxhinDB:
Thanks for this @dario23 -- merging.

PR #393, opened 2019-12-21
Enforce safety in checksum functions

### LINKED ISSUE #392: Possible UB in pnet_packet::util::checksum
util::checksum, a public function, calls sum_be_words which converts a &[u8] to a &[u16], however it does not properly check alignment before doing so which can lead to undefined behavior because the user may call the function with non-u16 aligned data.

TyPR124:
Fixes #392 by adding an assertion in the public `checksum` function. Also adds a debug assertion in the private `sum_be_bytes` to ensure it is not being used unsafely.

If a possible panic is undesired, then there may be another way to fix this. This way seems least likely to have a noticeable performance impact (but I haven't tested the performance impact).

The short-circuit for `data.len() == 0` is necessary because an empty slice has address 1, which would fail the assertion. Technically a short-circuit could also be added for `data.len() == 1` though I have not done so.

### DIFF
--- a/pnet_packet/src/util.rs
+++ b/pnet_packet/src/util.rs
@@ -64,8 +64,11 @@ impl Octets for u8 {
 }
 
 /// Calculates a checksum. Used by ipv4 and icmp. The two bytes starting at `skipword * 2` will be
-/// ignored. Supposed to be the checksum field, which is regarded as zero during calculation.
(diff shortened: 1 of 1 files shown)


JuxhinDB:
Hey @TyPR124 -- thanks a lot for this. Taking a look at this, I don't see any issues. Going to go ahead and merge.

PR #394, opened 2020-01-03
Fix a bug in calling `pselect`

zhangxp1998:
Per `pselect` [man pages](https://developer.apple.com/library/archive/documentation/System/Conceptual/ManPages_iPhoneOS/man2/select.2.html#//apple_ref/doc/man/2/select), the `k`th bit of `self.fd_set` must be set if we are interested in file descriptor `k`. libpnet does this once in [here](https://github.com/libpnet/libpnet/blob/dc2203a32a00981995b548eed4b4a90010e9491e/pnet_datalink/src/bpf.rs#L232), when the datalink receiver/sender is initialized. However, if no descriptors are available during a `pselect` call, the kernel will clear `self.fd_set`, and next time when pnet calls `pselect`, it will pass a `fd_set` with all bits 0. The kernel will interpret this as "I don't care if any of the file descriptors are available", and will happily return 0 even if there are some packets available. See screenshot, all bits in the `fd_set` bit mask are set to 0(this is not supposed to happen!).

Xnip2020-01-03_16-46-22

This caused an issue in another rust open source project, see discussion [here](https://github.com/imsnif/bandwhich/issues/51#issuecomment-570704211).

### DIFF
--- a/pnet_datalink/src/bpf.rs
+++ b/pnet_datalink/src/bpf.rs
@@ -260,6 +260,7 @@ impl DataLinkSender for DataLinkSenderImpl {
             for chunk in self.write_buffer[..len].chunks_mut(packet_size) {
                 func(chunk);
                 let ret = unsafe {
+                    libc::FD_SET(self.fd.fd, &mut self.fd_set as *mut libc::fd_set);
                     libc::pselect(self.fd.fd + 1,
                                   ptr::null_mut(),
                                   &mut self.fd_set as *mut libc::fd_set,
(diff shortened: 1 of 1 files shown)


JuxhinDB:
Apologies for the late reply from our end. Looking into this now and will update you shortly.

zhangxp1998:
For reference, linux's implementation sets the bit mask before every `pselect` call:
https://github.com/libpnet/libpnet/blob/dc2203a32a00981995b548eed4b4a90010e9491e/pnet_datalink/src/linux.rs#L248

JuxhinDB:
This is a nice catch, thanks a lot for fixing it @zhangxp1998. I just went over the diff and have no issues with this. Will be merging shortly :-)

Do you require a version bump?

zhangxp1998:
> Do you require a version bump?

Yes please!

JuxhinDB:
@zhangxp1998 -- version bump in progress. Bear with me if I'm a tad slow. :-)

JuxhinDB:
@zhangxp1998 Version has been bumped to 0.25.0

zhangxp1998:
> @zhangxp1998 Version has been bumped to 0.25.0

Awesome! Thanks a lot!

Issue #381, opened 2019-08-06
No packets captured on local interface


dantodor:
I'm trying to run the packetdump example on local interface to see packet capture capabilities. 
Context: macOs 10.14.5 Mojave, rustc stable 1.36
While there are no errors and no compilation issues, the output I get on lo0 is 

`[lo0]: Unknown packet: 00:00:00:00:00:00 > 00:00:00:00:00:00; ethertype: EtherType(0) length: 66

[lo0]: Unknown packet: 00:00:00:00:00:00 > 00:00:00:00:00:00; ethertype: EtherType(0) length: 66

[lo0]: Unknown packet: 00:00:00:00:00:00 > 00:00:00:00:00:00; ethertype: EtherType(0) length: 66

[lo0]: Unknown packet: 00:00:00:00:00:00 > 00:00:00:00:00:00; ethertype: EtherType(0) length: 66

[lo0]: Unknown packet: 00:00:00:00:00:00 > 00:00:00:00:00:00; ethertype: EtherType(0) length: 66

[lo0]: Unknown packet: 00:00:00:00:00:00 > 00:00:00:00:00:00; ethertype: EtherType(0) length: 66

[lo0]: Unknown packet: 00:00:00:00:00:00 > 00:00:00:00:00:00; ethertype: EtherType(0) length: 66

[lo0]: Unknown packet: 00:00:00:00:00:00 > 00:00:00:00:00:00; ethertype: EtherType(0) length: 66

[lo0]: Unknown packet: 00:00:00:00:00:00 > 00:00:00:00:00:00; ethertype: EtherType(0) length: 66

[lo0]: Unknown packet: 00:00:00:00:00:00 > 00:00:00:00:00:00; ethertype: EtherType(0) length: 74

[lo0]: Unknown packet: 00:00:00:00:00:00 > 00:00:00:00:00:00; ethertype: EtherType(0) length: 86

[lo0]: Unknown packet: 00:00:00:00:00:00 > 00:00:00:00:00:00; ethertype: EtherType(0) length: 74

[lo0]: Unknown packet: 00:00:00:00:00:00 > 00:00:00:00:00:00; ethertype: EtherType(0) length: 86

[lo0]: Unknown packet: 00:00:00:00:00:00 > 00:00:00:00:00:00; ethertype: EtherType(0) length: 74

[lo0]: Unknown packet: 00:00:00:00:00:00 > 00:00:00:00:00:00; ethertype: EtherType(0) length: 86

[...]

PR #395, opened 2020-01-04
Fix packet dump example on macos loopback interface.

hughesac:
This PR fixes https://github.com/libpnet/libpnet/issues/381.  The problem was that the BPF datalink code adds a zeroed out Ethernet header to the beginning of the packet when the interface is loopback.  None of the code in the packetdump example code was handling this case.


JuxhinDB:
Thanks for this @hughesac! Going to go ahead and merge :-)

Issue #399, opened 2020-01-10
pnet_packet 0.25 panic in pnet_packet::ipv4::checksum


chifflier:
Upgrade to 0.25 broke things, for ex I got a panic in ipv4::checksum:
```
thread 'main' panicked at 'assertion failed: `(left == right)`
  left: `0`,                                                                                                                                                           
 right: `1`: Cannot sum mis-aligned words at 0x7b2145203865', /home/pollux/.cargo/registry/src/github.com-1ecc6299db9ec823/pnet_packet-0.25.0/src/util.rs:71:5
stack backtrace:         
   0: backtrace::backtrace::libunwind::trace 
             at /cargo/registry/src/github.com-1ecc6299db9ec823/backtrace-0.3.40/src/backtrace/libunwind.rs:88
   1: backtrace::backtrace::trace_unsynchronized                                                                                                                       
             at /cargo/registry/src/github.com-1ecc6299db9ec823/backtrace-0.3.40/src/backtrace/mod.rs:66
   2: std::sys_common::backtrace::_print_fmt
             at src/libstd/sys_common/backtrace.rs:77
   3: <std::sys_common::backtrace::_print::DisplayBacktrace as core::fmt::Display>::fmt
             at src/libstd/sys_common/backtrace.rs:61
   4: core::fmt::write                                                             
             at src/libcore/fmt/mod.rs:1028
   5: std::io::Write::write_fmt                                                    
             at src/libstd/io/mod.rs:1412 
   6: std::sys_common::backtrace::_print
             at src/libstd/sys_common/backtrace.rs:65
   7: std::sys_common::backtrace::print
             at src/libstd/sys_common/backtrace.rs:50
   8: std::panicking::default_hook::{{closure}}                                                                                                                        
             at src/libstd/panicking.rs:188
   9: std::panicking::default_hook
             at src/libstd/panicking.rs:205
  10: std::panicking::rust_panic_with_hook                                                                                                                             
             at src/libstd/panicking.rs:464                   
  11: std::panicking::continue_panic_fmt
             at src/libstd/panicking.rs:373
  12: std::panicking::begin_panic_fmt
             at src/libstd/panicking.rs:328
  13: pnet_packet::util::checksum
             at /home/pollux/.cargo/registry/src/github.com-1ecc6299db9ec823/pnet_packet-0.25.0/src/util.rs:71
  14: pnet_packet::ipv4::checksum
[...]

PR #403, opened 2020-01-17
Fix issue 399 by using u16::from_be_bytes so we aren't using any raw …

hughesac:
This fixes the panic (#399) that was happening when the memory used in checksum calculations was not 16-bit aligned.  The code no longer uses raw pointers so we don't need any checks for alignment.  I ran the pnet_packet unit tests.

### DIFF
--- a/pnet_packet/src/util.rs
+++ b/pnet_packet/src/util.rs
@@ -170,6 +170,7 @@ fn sum_be_words(data: &[u8], skipword: usize) -> u32 {
 #[cfg(test)]
 mod tests {
     use super::sum_be_words;
+    use std::slice;
 
     #[test]
     fn sum_be_words_different_skipwords() {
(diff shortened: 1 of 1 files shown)


JuxhinDB:
@mrmonday I would really love your thoughts on this one (or @imsnif should you want to pitch in as it affects you).

mrmonday:
I'm in favour of anything that gets rid of unsafe code.

Would be good to get a new unit test for the failing case so we don't regress.

IIRC, the checksum code has been redone a bunch of times to make it faster and fix bugs - so bear that in mind.

JuxhinDB:
Thanks for the input @mrmonday!

@hughesac thanks for the PR (apologies for not mentioning this earlier). Following up on @mrmonday comment, would we be able to include some [regression] tests to sustain the PR?

hughesac:
Yes I should be able to add some tests within the next couple days.

JuxhinDB:
Looking forward. Will keep an eye out for changes.

hughesac:
I've finished adding tests.  Let me know if there are others you would like to see.

JuxhinDB:
Went through the changes and tests added. Apologies for taking a while to follow-up on this. Thanks @hughesac, merging!

hughesac:
@JuxhinDB  Can you make a new release that includes these commits so that dependent projects can get them?

JuxhinDB:
Hey @hughesac -- I set a reminder to do so. But realistically we're looking at mid next week, apologies

PR #408, opened 2020-01-24
Send the packet using the port specified by the user

dalurness:
Allows user to specify which port they would like their packet to be sent from.


pheki:
Shouldn't the port number in examples and tests be `0` instead of `12345`? Note that port 0 usually means any available, dynamically allocated port.

dalurness:
> Shouldn't the port number in examples and tests be `0` instead of `12345`? Note that port 0 usually means any available, dynamically allocated port.

 @pheki Yeah, port 0 is commonly used for tests and in many use cases for port management. I'll switch it over in the test cases. However, it doesn't really matter which port is used, as long as it doesn't use a reserved port or a port that has already been bound.

JuxhinDB:
Hi @dalurness, thanks a lot for the PR. Will be looking into it properly in coming days. Apart from adjusting the API to now include the port, everything else seems to be pretty standard. Would require a version bump.

kishiguro:
we may approach to this with Builder Pattern with set_port(u16) to TransportSender.  let me try create a PR with it.

PR #426, opened 2020-04-27
Update ipnetwork from 0.15.0 to 1.16.0.

kishiguro:


kishiguro:
We can use original #401 for merge.

PR #436, opened 2020-04-29
Enable pnet to build on Apple iOS target

thombles:
Right now there are various mods that go undefined since it doesn't match any of the existing `target_os` types.

Solution: treat it as equivalent to `macos` in all cases.

Really I'm only using `datalink::interfaces()` but that much is working fine on my iPad.

### DIFF
--- a/examples/packetdump.rs
+++ b/examples/packetdump.rs
@@ -249,7 +249,7 @@ fn main() {
         match rx.next() {
             Ok(packet) => {
                 let payload_offset;
-                if cfg!(target_os = "macos")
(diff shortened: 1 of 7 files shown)


kishiguro:
sometimes there is a travis-ci error on test layer2 with timeout.   let me take a look into the issue.  since this pr is not related to the travis-ci error, let me merge it.

thombles:
Thanks for looking at these so quickly - much appreciated!

kishiguro:
Thanks for contributing libpnet project!

PR #437, opened 2020-04-29
Update regex dependency from 1.1 to 1.3

thombles:
For compatibility with projects using the newer version


kishiguro:
thanks, let me merge it.

PR #439, opened 2020-05-04
Add LICENSE file and readme infomation in Cargo.toml.

kishiguro:


mrmonday:
This would probably be nicer as symlinks, rather than copying the files.

kishiguro:
thanks for suggestion.  made the license files to symlinks.

PR #441, opened 2020-05-08
Enable promiscuous capture for pcap

masche842:
Promiscuous mode is activated (not configurable) in the native unix implementation of libpnet, but disabled when using pcap.
I think it would be better to unify behaviour for all platforms/implementations. The easiest way is to activate promiscuous mode in pcap, which done in this pull request.
A more advanced step would be to make it configurable, e.g. via the config object. While I think the former makes sense in any case I'd like to know what others think about the latter.


kishiguro:
agreed, let's make the behavior consistent and have a configuration for the feature later on.

PR #445, opened 2020-05-15
Fix calculation of pcap read timeout

masche842:
Calculation of pcap timeout is faulty at the moment (as_subsec_nanos / 1000 yields microseconds, not milliseconds as intended). 
This is fixed in this pull request and as_subsec_nanos is replaced with as_subsec_millis method (which didn't exist back then).
As of rust 1.33, an as_millis method exists as well, but it returns u128, so it might have a small performance impact (though we haven't tested it)


kishiguro:
Thanks for the fix.  Let me merge it.

PR #455, opened 2020-07-27
Fix undefined behavior in packet field setters

### LINKED ISSUE #411: set_payload_length not labeled unsafe and does not check length of buffer
MWE:

```rust
use pnet::packet::udp::MutableUdpPacket;

fn main() {
    let mut pkt = vec![0; 32];
    let mut eth = MutableUdpPacket::new(&mut pkt).unwrap();
    let bad_dat = vec![0; 10000];
    eth.set_payload(bad_dat.as_slice());
}
```

All safe rust but results in a heap overflow.

Miri output:

```
error: Miri evaluation error: Memory access failed: pointer must be in-bounds at offset 10008, but is outside bounds of allocation 530 which has size 32
    -- /home/.rustup/toolchains/nightly-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/src/libcore/intrinsics.rs:1523:5
     |
1523 |     copy_nonoverlapping(src, dst, count)
     |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ Memory access failed: pointer must be in-bounds at offset 10008, but is outside bounds of allocation 530 which has size 32
     |
     = note: inside call to std::intrinsics::copy_nonoverlapping::<u8> at /home/mwe/target/debug/build/pnet_packet-6fcf04b13261d69f/out/udp.rs:294:13
note: inside call to pnet_packet::udp::MutableUdpPacket::set_payload at src/main.rs:7:5
    -- src/main.rs:7:5
     |
7    |     eth.set_payload(bad_dat.as_slice());
     |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
     =

error: aborting due to previous error
```

Checking the length of a payload write would be a decent first step for dealing with this and support in-place modification. Ideally, though I think it might be reasonable for MutablePackets to operate over https://docs.rs/bytes/0.5.4/bytes/trait.BufMut.html or BytesMut for that matter. Would allow for easier packet manipulation.

### LINKED ISSUE #449: Segfault in icmp send
https://github.com/0xd34b33f/cloudflare-ping/blob/06e359d16fff09839cc6a09b1ebdcacb9788ee91/src/ping_core.rs#L332
Segfault, when sending icmp packet. Works in debug mode or in --march=native. Doesn't work in release mode

tdryer:
It is currently possible to trigger undefined behavior in field setters
created by the `packet` macro, such as `MutableIpv4Packet::set_payload`,
by setting a buffer larger than the remaining space in the packet
buffer.

Fix this by replacing the unsafe `copy_nonoverlapping` call with
`copy_from_slice`. This function still compiles to `memcpy`, but
includes a range check which which will panic if the destination buffer
is too small.

Fixes #449. Fixes #411.

### DIFF
--- a/pnet_macros/src/decorator.rs
+++ b/pnet_macros/src/decorator.rs
@@ -657,14 +657,10 @@ fn handle_vec_primitive(
         };
 
         let copy_vals = if inner_ty_str == "u8" {
-            // Efficient copy_nonoverlapping (memcpy)
+            // Efficient copy_from_slice (memcpy)
             format!("
-                                    // &mut and & can never overlap
-                                    unsafe {{
-                                        copy_nonoverlapping(vals[..].as_ptr(),
-                                                            _self.packet[current_offset..].as_mut_ptr(),
-                                                            vals.len())
-                                    }}
+                                    _self.packet[current_offset..current_offset + vals.len()]
+                                        .copy_from_slice(vals);
                                 ")
         } else {
             // e.g. Vec<u16> -> Vec<u8>
@@ -689,7 +685,6 @@ fn handle_vec_primitive(
                                 #[allow(trivial_numeric_casts)]
                                 #[cfg_attr(feature = \"clippy\", allow(used_underscore_binding))]
                                 pub fn set_{name}(&mut self, vals: &[{inner_ty_str}]) {{
-                                    use std::ptr::copy_nonoverlapping;
                                     let mut _self = self;
                                     let current_offset = {co};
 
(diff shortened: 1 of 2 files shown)


kishiguro:
CI failed with failed to load source for dependency `ipnetwork`...
Let me try to re-run CI for figuring out root cause.

mrmonday:
Just rebased this, hopefully CI will pass now.

mrmonday:
CI is taking too long, gonna merge this and test it locally.

PR #456, opened 2020-08-07
Improve datalink interfaces example

zaszi:
This is a very minor documentation improvement that replaces the example code of the datalink interfaces() function with more idiomatic Rust.


kishiguro:
Thanks for contributing.  Just merged.

PR #459, opened 2020-08-31
Add description field to network interface (winpcap)

### LINKED ISSUE #348: Windows: provide NetworkInterface description in addition to name
Windows/winpcap has the IP_ADAPTER_INFO type and PacketGetAdapterNames function to enumerate network interfaces. While the names are not very human-readable (e.g. F48A60DC-4371-4970-B91A-88C8773E1452 and \Device\NPF_{E6BA643C-B5C1-4A7A-9A1F-900312083BBD}, respectively), IP_ADAPTER_INFO->Description is much better: Realtek PCIe GBE Family Controller #2

pnet's NetworkInterface struct only has a name field, so the description cannot be gathered along with the name. There are two options to improve user experience (e.g. when designing a GUI): a) add a description field. downside: It will only be filled on Windows and stays empty on unix machines. b) add a method to NetworkInterface that returns the description. Same downside as a), but no need to change the struct.

Sequal32:
Simple addition, closes #348.

### DIFF
--- a/pnet_datalink/src/dummy.rs
+++ b/pnet_datalink/src/dummy.rs
@@ -167,6 +167,7 @@ pub fn interfaces() -> Vec<NetworkInterface> {
 pub fn dummy_interface(i: u8) -> NetworkInterface {
     NetworkInterface {
         name: format!("eth{}", i),
+        description: "".to_string(),
         index: i as u32,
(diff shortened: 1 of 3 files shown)


kishiguro:
Thanks for the PR, just merged.  Sorry for late response.

PR #488, opened 2021-05-09
add FromStr to IpNextHeaderProtocol + refactoring

MikHulk:
I need to retieve the proper protocol from a string so I have made this PR in order to implement FromStr on NextHeaderProtocol. In order to define only once the protocol list, this is done with macro.

### DIFF
--- a/pnet_packet/src/ip.rs
+++ b/pnet_packet/src/ip.rs
@@ -26,8 +26,8 @@ impl PrimitiveValues for IpNextHeaderProtocol {
     }
 }
 
-/// Build Const and implements Display and FromStr from this list of protocols
-/// Protocol numbers as defined at:
+/// Build Const and implements Display and FromStr from a list of protocols
+/// Protocol numbers are defined at:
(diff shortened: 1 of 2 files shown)


mrmonday:
I think I would rather see a non-macro based implementation for this.

Yes, it's repetitive, but it's also a lot easier to read.

MikHulk:
Indeed the macro hide the real implementation. On the other hand it's a Display and FromStr implementations and the protocole and their numbers are given. The real issue with macro  is that constants are hidden too. But I understand your point of view. It was a good exercise for me (I am learning macro right now). In the same time  I have found difficult to get the proper constant in the lib from an user input. I already had this issue with other constants in pnet. Usually I make a match clause in my own code but in this case (255 constants) I have found that it would lt be nice if we could have a FromStr implementation in the lib directly.

MikHulk:
I see 2 possibilities:

* macro for Display / FromStr keeping the constants explicitely,
* an explicit implementation of FromStr like the Display implementation, (it's not hard to do with a search and replace regex

mrmonday:
Closing due to inactivity.

PR #490, opened 2021-05-09
Set up github actions

mrmonday:
Attempt to set up github actions. If it works I'll remove travis/appveyor.


mrmonday:
@chifflier any chance you could look at the macOS build failure here?

Looks like it is caused by `compiletest-rs`, it only happens on macOS though. I had to upgrade from `0.6` -> `0.7` to get around the DYLD issue we were seeing previously, otherwise this is the same as when you worked on it.

I've poked it a bit, but haven't had much luck.

chifflier:
> @chifflier any chance you could look at the macOS build failure here?
> 
> Looks like it is caused by `compiletest-rs`, it only happens on macOS though. I had to upgrade from `0.6` -> `0.7` to get around the DYLD issue we were seeing previously, otherwise this is the same as when you worked on it.
> 

I'll investigate
Note: in one of my other crates, I had to switch to `trybuild` because of similar reasons. Wondering if it could be a solution

chifflier:
I tried to debug the tests failure, and can confirm that
- I do not get why `compiletest_rs` does not find `pnet_macros_support`, while it is the current crate. This may be related to a wrong `DYLD_LIBRARY_PATH` value, but I'm not sure of that
- I tried using `trybuild` on the `macos-latest` box, and all tests succeeded. There is no difference in functionalities (and in fact, it is much easier to use `trybuild`).

@mrmonday would you agree to a PR switching to `trybuild` ? I can prepare that shortly, and I believe it would make the tests more reliable

mrmonday:
Thanks!

I've merged your changes and rebased this branch, lets see how it goes.

PR #491, opened 2021-05-10
pnet_packet: switch tests to trybuild

chifflier:
Use `trybuild` instead of `compiletest_rs` for `pnet_macros` compile tests.

The only difference is that the expected errors are now committed to the repository (`trybuild` compares them to the actual output of the compiler).

This fixes the build failure on macos seen in #490 and should make compile tests more reliable on such targets.


mrmonday:
Nice! Looks good, thank you.

PR #500, opened 2021-05-24
module path should be pnet::datalink

hsmtkk:
I found a document error in lib.rs.
A module path should be pnet::datalink, not pnet_datalink.


stappersg (approved):
Looks good to me

mrmonday:
Thanks!

PR #503, opened 2021-05-30
example/packetdump handle ethernet packet anyway

stappersg:
The check on Apple looks strange for non Apple users.
It took me a while to understand that `handle_ethernet_frame()`
even happens twice on Apple systems.

This change aims for making the example more self explaining.


mrmonday:
This change breaks the example on `macos` and `ios`.

There shouldn't need to be special handling for either platform - it seems there's a bug in `pnet_datalink`.

Unfortunately I don't have a mac to test this on, perhaps someone who does could look into this.

PR #504, opened 2021-05-30
Handle ethernet frame only once on Apple

stappersg:
After the "if MacOS or iOS then handle_ethernet_frame()"
there was an unconditional handle_ethernet_frame().

Now we have

   if MacOS or iOS {
       special_actions_for_Apple_OS();
       handle_ethernet_frame();
   } else {
       handle_ethernet_frame();
   }

to prevent triggering  handle_ethernet_frame() on Apple OS
without the special actions.

### DIFF
--- a/examples/packetdump.rs
+++ b/examples/packetdump.rs
@@ -282,8 +282,9 @@ fn main() {
                             continue;
                         }
                     }
+                } else {
+                    handle_ethernet_frame(&interface, &EthernetPacket::new(packet).unwrap());
                 }
-                handle_ethernet_frame(&interface, &EthernetPacket::new(packet).unwrap());
             }
             Err(e) => panic!("packetdump: unable to receive packet: {}", e),
         }


mrmonday:
Unfortunately this will break macOS too.

In the `macos` specific `if`, there is a [`continue`](https://doc.rust-lang.org/reference/expressions/loop-expr.html?highlight=continue#continue-expressions) after each `handle_ethernet_frame()`. This causes control to flow to the next iteration of the loop - it doesn't call `handle_ethernet_frame()` twice.

Adding in this `else` can cause `handle_ethernet_frame()` to not be called in some scenarios on `macos` and `ios`, leading to missed packets.

stappersg:
Ah, yes. Now I see the [`continue`](https://github.com/libpnet/libpnet/blob/master/examples/packetdump.rs#L275) and the [`continue`](https://github.com/libpnet/libpnet/blob/master/examples/packetdump.rs#L282).

Thanks for elaborating.

Issue #481, opened 2021-04-09
Packetdump example not working on windows


abusalaa7:
I was able to run the [packet dump example](https://github.com/libpnet/libpnet/blob/master/examples/packetdump.rs) on macos and on ubuntu but on windows it doesn't seem to capture any packets. The program is running but it's not printing any output and not throwing any errors or exiting.

I followed the instructions and installed winpcap 4.1.3 and copied the dll to this directory
\.rustup\toolchains\stable-x86_64-pc-windows-msvc\lib\rustlib\x86_64-pc-windows-msvc\lib

I also tried uninstalling winpcap, installing npcap with the api compatibility mode. I left the dll in the lib folder.

The program compiles and my printlln prints that it enters the loop only once. It seems like the rx is waiting forever without receiving any packets and never re-enters the loop.

Any help is appreciated.

Rust version: rustc 1.51.0 (2fd73fabe 2021-03-23)
Windows: 10 version 1803

StanislavGlebik:
Run into the same issue with arp example

ridgebacktom:
examples/packetdump.rs seems to have code that is not for macos or ios only:
https://github.com/libpnet/libpnet/blob/2ad90983ddc10e985392872c6ffd1e3fef7ce2d5/examples/packetdump.rs#L252

Was the above intentional? Are the examples meant to be cross-platform?

abusalaa7:
I tried commenting that part and it still won't work.

ridgebacktom:
On Windows, using nightly, it seems like the offset() function is not correctly setting the end pointer:
https://github.com/libpnet/libpnet/blob/2ad90983ddc10e985392872c6ffd1e3fef7ce2d5/pnet_datalink/src/winpcap.rs#L246

Inserting a debug, I see that **buflen** is continually changing, reflecting the capture of frames, but when **ptr** and **end** are set, the **end** pointer is always the same as the **ptr** variable.
https://github.com/libpnet/libpnet/blob/2ad90983ddc10e985392872c6ffd1e3fef7ce2d5/pnet_datalink/src/winpcap.rs#L241

The offset() function seems to be an experimental function on the rustc nightly toolchain.
https://doc.rust-lang.org/std/intrinsics/fn.offset.html

I did some experiments, and it seems that neither add() nor offset() (pointer functions, see https://doc.rust-lang.org/std/primitive.pointer.html#method.add) are working the way I expect. For example, the following code prints the same value for both pointers:

```
unsafe {
[...]

PR #506, opened 2021-06-14
Edited windows requirements

abusalaa7:
Use winapi in cargo file to fix the issue in #481

### DIFF
--- a/README.md
+++ b/README.md
@@ -89,3 +89,8 @@ There are three requirements for building on Windows:
    in a directory named `lib`, in the root of this repository. Alternatively, you can use any of the
    locations listed in the `%LIB%`/`$Env:LIB` environment variables. For the 64 bit toolchain it is
    in `WpdPack/Lib/x64/Packet.lib`, for the 32 bit toolchain, it is in `WpdPack/Lib/Packet.lib`.
+ * You must add winapi to the cargo file
+   ```
+   [target.'cfg(windows)'.dependencies]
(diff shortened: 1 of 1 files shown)


mrmonday:
Why is this necessary?

It should be picked up transitively through libpnet's dependencies, no?

stappersg:
> It should be picked up transitively through libpnet's dependencies, no?

In #481 were two complians of brokeness on MS Windows. A third person offered a solution.

That solution resulted in this merge request.


I think the question _"Why is this necessary?"_ should be left unanswered.

What I further think:
* The patch should be against `Cargo.toml`
* The patch should be accepted
* The question should be _"Whom is harm by this proposed change?"_

mrmonday:
Ah, found it:

https://github.com/libpnet/libpnet/issues/481#issuecomment-834821390

I think the real fix for this would be to change the `winapi` dependency in libpnet to include that feature - that should cause it to "just work" without any manual intervention from users.

I won't have a Windows machine to test this on.

abusalaa7:
I have a windows machine, I will test it.

abusalaa7:
@mrmonday @stappersg - noticed something weird. Building from source on windows, without any changes to `cargo.toml` file doesn't cause the same issue. I can still replicate the issue when using the crate but if I build the source on windows and reference its location on disk, it works fine. I can't confirm if adding `winapi` dependency to cargo file will fix it. Any suggestions?

mrmonday:
Closing due to inactivity.

Issue #513, opened 2021-07-22
datalink::DataLinkSender::build_and_send() and channel's write_buffer


swiryaman:
The check on the linux version of `build_and_send()` for the length of the packets and the write_buffer size is off by one byte.  I thinks it should be `<=` instead of `<` in https://github.com/libpnet/libpnet/blob/a7b79d85f59a7f8e41d06fda772d278630fcfde6/pnet_datalink/src/linux.rs#L245

stappersg:
On Thu, Jul 22, 2021 at 07:07:16AM -0700, Santa Wiryaman wrote:
> thinks it should be

Patches welcome

swiryaman:
Just a one liner change.  Without this change, build_and_sending 4 1024-byte packets retturns `None` even though the default write_buffer_size is 4096 bytes.

```
diff --git a/pnet_datalink/src/linux.rs b/pnet_datalink/src/linux.rs
index 4ec7a96..0fe9161 100644
--- a/pnet_datalink/src/linux.rs
+++ b/pnet_datalink/src/linux.rs
@@ -242,7 +242,7 @@ impl DataLinkSender for DataLinkSenderImpl {
         func: &mut dyn FnMut(&mut [u8]),
     ) -> Option<io::Result<()>> {
         let len = num_packets * packet_size;
-        if len < self.write_buffer.len() {
+        if len <= self.write_buffer.len() {
             let min = cmp::min(self.write_buffer[..].len(), len);
             let mut_slice = &mut self.write_buffer;
             for chunk in mut_slice[..min].chunks_mut(packet_size) {
```

PR #515, opened 2021-07-27
 Fix length check in datalink's build_and_send()

swiryaman:
I reported Issue #513.  Here is the simple fix for it.


mrmonday:
Looks good, thanks!

PR #529, opened 2021-10-27
Mark pnet_datalink::Channel non-exhaustive; remove workaround

cole-miller:
`pnet_datalink::Channel` has a dummy `PleaseIncludeACatchAllVariantWhenMatchingOnThisEnum` variant, with the stated goal of allowing the library to add more variants in the future without breaking existing code. This is the purpose of Rust's [`non_exhaustive` attribute](https://doc.rust-lang.org/stable/reference/attributes/type_system.html#the-non_exhaustive-attribute), so remove the dummy variant and mark `pnet_datalink::Channel` as non-exhaustive instead.

I also fixed some ill-formed `panic!` invocations in the examples.

### DIFF
--- a/examples/fanout.rs
+++ b/examples/fanout.rs
@@ -82,7 +82,7 @@ fn main() {
                 // Create a channel to receive on
                 let (_, mut rx) = match datalink::channel(&itf, config) {
                     Ok(Ethernet(tx, rx)) => (tx, rx),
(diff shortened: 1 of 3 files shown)


mrmonday:
Thank you for this - this code was written before `#[non_exhaustive]` existed.

cole-miller:
Thanks!

PR #534, opened 2021-11-30
Add NetworkInterface::is_running()

Disasm:
This PR adds `is_running` method aimed to detect the state of the network interface link. It was tested on my x86_64 Linux machine by running the `list_interfaces` example and observing its output when the network cable was connected/disconnected.

### DIFF
--- a/pnet_datalink/src/lib.rs
+++ b/pnet_datalink/src/lib.rs
@@ -245,6 +245,11 @@ impl NetworkInterface {
     pub fn is_up(&self) -> bool {
         self.flags & (pnet_sys::IFF_UP as u32) != 0
     }
+
+    pub fn is_running(&self) -> bool {
+        self.flags & (pnet_sys::IFF_RUNNING as u32) != 0
+    }
+
     pub fn is_broadcast(&self) -> bool {
         self.flags & (pnet_sys::IFF_BROADCAST as u32) != 0
     }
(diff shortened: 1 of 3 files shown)


stappersg:
I do miss the _why_.

And I don't know when the Linux kernel does set the `RUNNING` flag.
My observation is that `ip` doesn't care about it.
```text
$ ip link show dev lo
1: lo: <LOOPBACK,UP,LOWER_UP> mtu 65536 qdisc noqueue state UNKNOWN mode DEFAULT group default qlen 1000
    link/loopback 00:00:00:00:00:00 brd 00:00:00:00:00:00
$ ip address show dev lo
1: lo: <LOOPBACK,UP,LOWER_UP> mtu 65536 qdisc noqueue state UNKNOWN group default qlen 1000
    link/loopback 00:00:00:00:00:00 brd 00:00:00:00:00:00
    inet 127.0.0.1/8 scope host lo
       valid_lft forever preferred_lft forever
    inet6 ::1/128 scope host 
       valid_lft forever preferred_lft forever
$ /sbin/ifconfig lo
lo: flags=73<UP,LOOPBACK,RUNNING>  mtu 65536
        inet 127.0.0.1  netmask 255.0.0.0
        inet6 ::1  prefixlen 128  scopeid 0x10<host>
        loop  txqueuelen 1000  (Local Loopback)
        RX packets 2553  bytes 236862 (231.3 KiB)
        RX errors 0  dropped 0  overruns 0  frame 0
        TX packets 2553  bytes 236862 (231.3 KiB)
        TX errors 0  dropped 0 overruns 0  carrier 0  collisions 0

$ 
```

@Disasm please enlighten us about the use case to the proposed function.

Disasm:
`RUNNING` is set when the network connection is ready for data transfer: cables are connected and connection speed is negotiated. This flag is useful when one needs to check that it makes sense to send data packets. `UP` only means that the network interface is activated, it tells nothing about the link status.

stappersg:
Mmm, the `RUNNING` of `ifconfig` is probably the `LOWER_UP` from `ip`.
Command `ip`  shows `LOWER_UP` upon _link detect".
```text
$ ip a | grep UP
1: lo: <LOOPBACK,UP,LOWER_UP> mtu 16436 qdisc noqueue state UNKNOWN group default 
2: eth0: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc pfifo_fast state UP group default qlen 1000
```
Thing I'm trying to share (besides _the merge request has been noticed_) is that `ifconfig` development didn't follow development of the Linux kernel. `ifconfig` did choose _works also outside Linux (think BSD and other Unixes) and works good enough_.
`ip` did choose _we follow each Linux network kernel development_. 
I switched also to `ip`.  I did not `RUNNING` for a long time, hence I started asking questions about it.

mrmonday:
I think this looks good, thanks.

Disasm:
Thank you!

PR #540, opened 2022-01-11
Added support for querying kernel driver flags

### LINKED ISSUE #538: Is is_up() working correctly? Linux
I have several interfaces on my machine. Many are internal and show "state DOWN" using the ip addr command. However, if I test against is_up() all of the interfaces return in the affirmative. The code below prints out all of the interfaces, not just those showing "state UP".

```
let ifaces = pnet::datalink::linux::interfaces();

for inf in ifaces{
    if inf.is_up(){
        println!("{}",inf);
    }
}
```

fennewald:
Closes #538

Adds support for the `IFF_LOWER_UP` and `IFF_DORMANT` flags.

### DIFF
--- a/pnet_datalink/src/lib.rs
+++ b/pnet_datalink/src/lib.rs
@@ -245,6 +245,16 @@ impl NetworkInterface {
     pub fn is_up(&self) -> bool {
         self.flags & (pnet_sys::IFF_UP as u32) != 0
     }
+    /// Triggered when the driver has signated netif_carrier_on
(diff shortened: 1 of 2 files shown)


mrmonday:
Looks like this breaks the build on Windows, unfortunately.

Could you also update the `Display` impl too please?

mrmonday:
Going to work on this myself, since I need to fix up the other branch anyway.

mrmonday:
Rebased/updated in #540 - thank you for your work.

PR #541, opened 2022-01-11
Remove a few usages of uninitialized, add safety comments

5225225:
I added some safety comments to the `interfaces` method, and removed some uninitialized (which is UB in basically every situation).

I didn't touch the bits that `cargo check` can't check, and I didn't run tests, I'll leave that up to CI.

One thing I'm concerned about is the `String::from_utf8_unchecked` which implies that all interface names on unix are *always* valid UTF-8. According to https://unix.stackexchange.com/questions/451368/allowed-chars-in-linux-network-interface-names this is incorrect, so what should we do here? Panic? Return an error? Ignore the interface? Use CString in NetworkInterface?

I think using CString in NetworkInterface is the best choice. That would be a breaking change, but it's for the best IMO.

### DIFF
--- a/pnet_datalink/src/unix_interfaces.rs
+++ b/pnet_datalink/src/unix_interfaces.rs
@@ -17,6 +17,7 @@ use libc;
 
 use std::ffi::{CStr, CString};
 use std::mem;
+use std::mem::MaybeUninit;
 use std::net::IpAddr;
 use std::os::raw::c_char;
 use std::str::from_utf8_unchecked;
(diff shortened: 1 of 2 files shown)


pnet_datalink/src/unix_interfaces.rs
```diff
@@ -33,54 +34,74 @@ pub fn interfaces() -> Vec<NetworkInterface> {
...
+
+    let mut addr = addrs;
+    while !addr.is_null() {
+        // Safety: We assume that addr is valid for the lifetime of this loop
+        // body, and is not mutated.
+        let addr_ref: &libc::ifaddrs = unsafe {&*addr};
+
+        let c_str = addr_ref.ifa_name as *const c_char;
+
+        // Safety: ifa_name is a null terminated interface name
```
mrmonday, line 56: For assertions like this it would be nice to have information about why we know this - here it might be good to have something like:
```rust
// See also: man 3 getifaddrs
```

pnet_datalink/src/unix_interfaces.rs
```diff
@@ -33,54 +34,74 @@ pub fn interfaces() -> Vec<NetworkInterface> {
...
+        let bytes = unsafe { CStr::from_ptr(c_str).to_bytes() };
+
+        // Safety: Interfaces on unix must be valid UTF-8
+        // TODO: Really? They *must* be UTF-8?
```
mrmonday, line 60: I suspect this is not the case, and that it is a bug. It should probably be `OsString`, but it depends what Windows/other OSes do here.

pnet_datalink/src/unix_interfaces.rs
```diff
@@ -33,54 +34,74 @@ pub fn interfaces() -> Vec<NetworkInterface> {
...
+            ifaces.push(ni);
         }
+
+        addr = addr_ref.ifa_next;
+    }
+
+    // Safety: addrs has been previously allocated through getifaddrs
+    unsafe {
         libc::freeifaddrs(addrs);
+    }
```
mrmonday, line 93: I think this is technically a bug too, the code could panic before hitting here, and `addrs` would not be freed. I'm not sure what best practise for this is - presumably we need a resource struct to wrap this and ensure proper cleanup.

pnet_datalink/src/unix_interfaces.rs
```diff
@@ -33,54 +34,74 @@ pub fn interfaces() -> Vec<NetworkInterface> {
...
 
-        for iface in &mut ifaces {
-            let name = CString::new(iface.name.as_bytes()).unwrap();
+    for iface in &mut ifaces {
+        let name = CString::new(iface.name.as_bytes()).unwrap();
```
mrmonday, line 96: As long as there's API breakage, this should be fixed too - this function should return a `Result<>` rather than unwrapping.

Actually, a bunch of fields in that struct should probably be `Option<>`, come to think of it...

pnet_sys/src/unix.rs
```diff
@@ -107,10 +107,10 @@ pub mod public {
     }
 
     fn make_in6_addr(segments: [u16; 8]) -> In6Addr {
-        #[allow(deprecated)]
-        let mut val: In6Addr = unsafe { mem::uninitialized() };
-        val.s6_addr = unsafe {
-            mem::transmute([
+        // Safety: We're transmuting an array of ints to an array of ints.
+        // There is no padding involved, and they must be the same size.
```
mrmonday, line 111: Is this definitely true on all architectures?

This code could actually be written without the `transmute()` - that's probably a better option.

mrmonday (approved):
Thank you for your work!

I've left a bunch of comments - I'm going to merge this now anyway and if you'd like to follow up in another PR with changes you can.

pnet_sys/src/unix.rs
5225225: Well, the size of `u16` is fixed at 2 bytes, and the min/max values are fixed too. So there's no *room* for padding. (If it was 2 bytes long but needed to be padded to 4 byte boundaries, it would count as a 4 byte long type).

We could take a `bytemuck` dependency and use that to transmute, or just do it manually with shifts in a loop and hope it gets optimised. I've not benchmarked the latter, it's *probably* fine.

pnet_datalink/src/unix_interfaces.rs
5225225: Yep, an `addrs` struct sounds good. That or use scopeguard/similar to manually free it on drop, but this code is complex enough as is, I'd rather like to move the linked list traversal outside of it.

PR #561, opened 2022-04-25
Fix build on architectures with unsigned char

pkubaj:
ARM and POWER are affected.

### DIFF
--- a/pnet_datalink/src/bpf.rs
+++ b/pnet_datalink/src/bpf.rs
@@ -129,7 +129,7 @@ pub fn channel(network_interface: &NetworkInterface, config: Config) -> io::Resu
     }
     let mut iface: bpf::ifreq = unsafe { mem::zeroed() };
     for (i, c) in network_interface.name.bytes().enumerate() {
-        iface.ifr_name[i] = c as i8;
+        iface.ifr_name[i] = c as libc::c_char;
     }
 
     let buflen = config.read_buffer_size as libc::c_uint;


stappersg:
(Summary:  the proposed patch did get human attention)
```text
-        iface.ifr_name[i] = c as i8;
+        iface.ifr_name[i] = c as libc::c_char;
```

It is the `libc::` that makes me wonder if that will block using `libpnet` in a `no_std` environment.

Would like to known if
```text
-        iface.ifr_name[i] = c as i8;
+        iface.ifr_name[i] = c as c_char;
```
or even
```text
-        iface.ifr_name[i] = c as i8;
+        iface.ifr_name[i] = c;
```
will compile (will work) on _ARM_ and/of _POWERPC_.

(Feel free to render this comment as _just noise_.)

pkubaj:
> (Summary: the proposed patch did get human attention)
> 
> ```
> -        iface.ifr_name[i] = c as i8;
> +        iface.ifr_name[i] = c as libc::c_char;
> ```
> 
> It is the `libc::` that makes me wonder if that will block using `libpnet` in a `no_std` environment.
> 
> Would like to known if
> 
> ```
> -        iface.ifr_name[i] = c as i8;
> +        iface.ifr_name[i] = c as c_char;
> ```

`error[E0412]: cannot find type `c_char` in this scope`

> 
> or even
> 
> ```
> -        iface.ifr_name[i] = c as i8;
> +        iface.ifr_name[i] = c;
> ```

That indeed seems to work.

> 
> will compile (will work) on _ARM_ and/of _POWERPC_.
> 
> (Feel free to render this comment as _just noise_.)

infosechoudini:
can we push a new commit that doesnt have `c_char`?

infosechoudini:
in the future, we'll look into other methods of building cross-os and cross-arch

mrmonday:
Looks good thanks!

Issue #133, opened 2015-12-04
Support #![no_std]


lilred:
A lot of people are writing operating systems kernels in Rust, and they'll be looking for a network stack. `pnet` could be a choice option.

Some thoughts:

* `nom` implements this as a feature on their package. Compiling with the `core` feature means that `#![no_std]` compilation is supported.
* It's probably reasonable to assume support for `core`, `alloc`, and `collections`.
* On the other hand, assuming support for `libc` is not going to go well. AFAIK It's pretty much an interface to `glibc`, which I wouldn't expect anyone to port to their OS.

mrmonday:
I definitely would like to see support for libpnet in the Rust-based OSes - unfortunately it's not as simply as supporting `#![no_std]` though. Some thoughts on the matter:

 * All the datalink backends are operating system specific, so by implementing a backend for <some OS>, you could support `#![no_std]` for it, depending on how it's implemented. Right now all backends require `libstd`, since it is available on all supported OSes. I guess the exception could be if you're attempting to target an embedded Linux - I'd love to hear more from someone working on that so we can see what libpnet can do for that environment.
* The transport-level stuff requires a posix-like API from the OS - it's likely easiest to just disable this completely when targetting new OSes.
 * pnet_macros does rely somewhat on std, and so, transitively, so does libpnet - it should be fairly simple to remove this dependency though.
 * libpnet isn't a network stack - it could be (and hopefully will be!) used to build a network stack though. I'd love to see some of the OS guys build a network stack atop of libpnet for their OSes, it would mean we've got an instantly cross-platform network stack - one that works on both their OS, and on Windows/Linux/OS X/FreeBSD/Whatever else libpnet ends up supporting (and, maybe even other Rust OSes!). I'm envisaging something like what's implemented in the [seastar framework](http://www.seastar-project.org/) or [mTCP](https://www.usenix.org/conference/nsdi14/technical-sessions/presentation/jeong) - we're in the great position of not having to support a legacy interface, so we can build in things like zero-copy from the start.

I'm happy to mentor anyone who wants to start building a network stack with libpnet, or who wants to get it working on their OS.

little-dude:
> it would mean we've got an instantly cross-platform network stack 
that would be pretty awesome !

lilred:
>I'm happy to mentor anyone who wants to start building a network stack with libpnet, or who wants to get it working on their OS.

Well, hello there :)

[...]

PR #566, opened 2022-05-13
Fix feature deps in top-level Cargo.toml blocking #[no_std] use

Felix-El:


Felix-El:
This is also required in order to achieve #133.

PR #578, opened 2022-07-14
Apply dependabot on all subfolders

matteyeux:
Hi !

I noticed #576 doesn't update `ipnetwork ` for  `pnet_datalink/cargo.toml`. That's because dependabot.yml is configured to only  update the cargo.toml in the root directory of the repo.

This PR allows dependabot to update all subprojects of this repo.


mrmonday:
Looks good, thanks!

I think there are a couple of directories missing - is it possible to use a wildcard instead?

PR #584, opened 2022-09-12
datalink(linux): add feature to pass the fd (socket) to ::channel()

Martichou:
This can be useful if the caller want to manipulate the socket differently at creating. 
For example setting a BPF filter on the socket before passing it in the channel:

```
datalink::linux::channel(interface, config, socket_fd)
```

Signed-off-by: Martin Andre <martin.andre@tessares.net>

### DIFF
--- a/pnet_datalink/src/lib.rs
+++ b/pnet_datalink/src/lib.rs
@@ -156,6 +156,9 @@ pub struct Config {
     pub linux_fanout: Option<FanoutOption>,
 
     pub promiscuous: bool,
+
+    /// Linux only: The socket's file descriptor that pnet will use
+    pub socket_fd: Option<i32>,
 }
 
 impl Default for Config {
@@ -169,6 +172,7 @@ impl Default for Config {
             bpf_fd_attempts: 1000,
             linux_fanout: None,
             promiscuous: true,
+            socket_fd: None,
         }
     }
 }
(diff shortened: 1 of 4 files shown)


pnet_datalink/src/lib.rs
```diff
@@ -181,8 +181,18 @@ impl Default for Config {
 /// When matching on the returned channel, make sure to include a catch-all so that code doesn't
 /// break when new channel types are added.
 #[inline]
-pub fn channel(network_interface: &NetworkInterface, configuration: Config) -> io::Result<Channel> {
-    backend::channel(network_interface, (&configuration).into())
+pub fn channel(
+    network_interface: &NetworkInterface,
+    configuration: Config,
+    #[cfg(feature = "psocket")]
+    socket: i32
```
mrmonday, line 188: I think this should be part of the `Config` struct, rather than an additional argument.

pnet_datalink/src/linux.rs
```diff
@@ -97,19 +97,25 @@ impl Default for Config {
...
+        super::ChannelType::Layer3(proto) => (libc::SOCK_DGRAM, proto as i32),
     };
-    let socket = unsafe { libc::socket(libc::AF_PACKET, typ, proto.to_be() as i32) };
-    if socket == -1 {
-        return Err(io::Error::last_os_error());
-    }
-    let mut addr: libc::sockaddr_storage = unsafe { mem::zeroed() };
-    let len = network_addr_to_sockaddr(network_interface, &mut addr, proto as i32);
 
+    #[cfg(not(feature = "psocket"))]
```
mrmonday, line 111: I don't think this should be feature gated - it should be configurable at runtime using the `Config` struct - if the socket is already set then we skip this setup.

mrmonday (changes requested):

Martichou:
Hey @mrmonday , sorry for the very very late reply and update on this. Just came back to this and still think this can be helpful to have in the lib.

The tests failing seems to not be linked with the change, can they be ignored? Or any idea?

mrmonday:
I've just rebased this. There are a couple of white space issues but if CI passes I'll fix that and get this merged in.

mrmonday:
Thanks! Sorry this took so long.

PR #586, opened 2022-10-01
Fix examples/arp_packet.rs

palindrom615:
IIUC, the goal of this code is to look up MAC address matching `target_ip`, by receiving ARP packet from the target host sends.

If so, catching subsequent packet won't work because the very next received packet is _the ARP request_ just sent by the function itself. (at least on my macbook)

To make this work, IMO, lookup every packets while reply packet found.


zszep:
Your fix made  the example run as expected. Got almost a headache wondering why the mac of my NIC was returned and not of the remote host. This should be merged into the main branch ASAP.

mrmonday:
Thanks!

PR #596, opened 2023-01-11
pnet_macros: Fix handle_vec_primitive for multi byte primitives

jefftt:
Currently `get_*()` will panic for primitives larger than 1 byte with an array out of bounds, e.g. `u32be` because it will try to push for `vec.capacity()` times which is set to the number of bytes in the vec. The correct times to push is the number of elements, which is the number of bytes divided by the size.


mrmonday:
Looks good, thanks!

PR #597, opened 2023-01-22
Updated example for version="0.31.0"

08r4y4n:


mrmonday:
I don't think these changes are needed since `v0.32.0` - could you check?

mrmonday:
Verified this locally - `pnet::datalink` not working was a regression from `nostd` support.

PR #599, opened 2023-02-01
Add blanket impls of Packet for Box<T> and &T.

landhb:
This is an ease-of-use improvement. Allowing users to pass boxed objects and references into `send_to` etc.

PR #606, opened 2023-02-03
Add blanket impls of Packet for Box<T> and &T.

landhb:
@mrmonday  Should fix #599 and #605


mrmonday:
Thanks for the quick turn around!

PR #625, opened 2023-07-10
TCP header remove NS flag

rikonaka:
ECN-nonce never gained widespread use and the RFC was moved to Historic status, so I remove it in TCP flags, and set flags in TCP header as u8 instead of u9be, set reserverd as u4 instead of u3.


rikonaka:
Hi @stappersg, I've resubmitted the code. 😜

mrmonday:
Thanks!

rikonaka:
> Thanks!

You are welcome 😜

PR #642, opened 2023-08-26
Update ARP example to also support IPv6 via NDP

tgross35:
This adds adds a NDP variant that handles ipv6 addresses.

I am not sure if there is a better way to get the intended MAC other than checking the ethernet packet's source. If there is, I can update.


mrmonday:
Thanks!

PR #647, opened 2023-09-21
Added a definition for ethernet flow control packets.

AJMansfield:
This is a pretty obsolete protocol and ended up not being useful for what I'd hoped it would be, but I wrote an implementation for it anyway so the world might as well have it.


AJMansfield:
(deleting in order to fix this so it merges from a feature branch)

PR #649, opened 2023-09-26
Definition for ethernet flow control packets.

AJMansfield:
This is a pretty obsolete protocol and ended up not being useful for what I'd hoped it would be, but I wrote an implementation for it anyway so the world might as well have it.

### DIFF
--- a/pnet_packet/src/flowcontrol.rs
+++ b/pnet_packet/src/flowcontrol.rs
@@ -1,0 +1,67 @@
+// Copyright (c) 2023 Anson Mansfield <amansfield@mantaro.com>
+//
+// Licensed under the Apache License, Version 2.0 <LICENSE-APACHE or
+// http://www.apache.org/licenses/LICENSE-2.0> or the MIT license
+// <LICENSE-MIT or http://opensource.org/licenses/MIT>, at your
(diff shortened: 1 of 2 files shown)


stevedoyle (approved):

stevedoyle (commented):
Missing unit test?

AJMansfield:
> Missing unit test?

AFAIK the standard doesn't give any test vectors, so I'm not sure how to meaningfully test this.

Inventing a test vector myself wouldn't add anything: if there _were_ an error, given the declarative nature of the code here the error would inevitably be a misinterpretation of the standard rather than a mis-implementation of that interpretation -- so adding a test vector I've concocted myself would at best amount to writing down the same mistake twice.

Would using Wireshark to create the test vector be appropriate here? Given that it's a much more mature project it might not be unreasonable to consider its behavior as "authoritative enough".

stevedoyle:
> Would using Wireshark to create the test vector be appropriate here? Given that it's a much more mature project it might not be unreasonable to consider its behavior as "authoritative enough".

I think using Wireshark or tcpdump to capture a reference/test packet is a good approach.

mrmonday:
Thanks!

PR #650, opened 2023-09-26
Add a next_with_timeout method to DataLinkReciever

AJMansfield:
One attempt at adding a way to do per-call timeouts on `DataLinkReciever`.
I added it this way for symmetry with the similarly-named `TransportChannel` methods, though I'm not sure if this is really the best way to do this; perhaps it would be better to just add a `set_timeout` method?

### DIFF
--- a/pnet_datalink/src/bpf.rs
+++ b/pnet_datalink/src/bpf.rs
@@ -432,6 +432,72 @@ impl DataLinkReceiver for DataLinkReceiverImpl {
         }
         Ok(&self.read_buffer[start..start + len])
     }
+
+    fn next_with_timeout<'a>(&'a mut self, t: Duration) -> io::Result<&[u8]> {
+        let timeout = Some(pnet_sys::duration_to_timespec(t));
+        // Loopback packets arrive with a 4 byte header instead of normal ethernet header.
+        // Discard that header and replace with zeroed out ethernet header.
+        let (header_size, buffer_offset) = if self.loopback {
+            (4, ETHERNET_HEADER_SIZE)
+        } else {
+            (0, 0)
+        };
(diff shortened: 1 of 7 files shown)


mrmonday:
I think a `set_timeout()` would be better, yes. Or possibly just an entry in the `Config` struct?

AJMansfield:
> I think a `set_timeout()` would be better, yes. Or possibly just an entry in the `Config` struct?

That does probably fit better with the design of the library; and I'll happily port my application code to use wrappers around that abstraction if it's added upstream instead of using my own fork.

Just to show how I'm using `next_with_timeout` from my fork, though, I use it specifically for a single-threaded multiplexed eventloop idiom (e.g. the type of eventloop calls like [`select(2)`](https://www.man7.org/linux/man-pages/man2/select.2.html) exist for in C). 
With this specific minimal example being to re-transmit packets after a delay:
```rust
loop {
    let rx_result = match tx_queue.do_delayed_retransmit(Instant::now()) // also returns tx time for the next packet after
            .map(|t| t.saturating_duration_since(Instant::now())) // how far in the future is that?
        { 
            Some(sleep) => rx.next_with_timeout(sleep), // sleep at most that long, don't miss the next tx time
            None => rx.next(), // sleep as long as possible, there's no next tx
        }.map(|pkt| (pkt, Instant::now())); // TODO: use SO_TIMESTAMP instead to get the packet timestamp from the kernel

    match rx_result {
        Ok((packet, now)) => tx_queue.schedule_delayed_retransmit(packet, now + delay),
        Err(ref err) if err.kind() == std::io::ErrorKind::TimedOut => {},
        Err(e) => panic!("Error recieving packet: {}", e),
    }
}
```

PR #652, opened 2023-10-10
Add apple tvos support

lcruz99:
This PR allows using libpnet lib to target [tier 3 *-apple builds](https://doc.rust-lang.org/nightly/rustc/platform-support/apple-tvos.html):

- Apple tvOS on aarch64
- Apple tvOS Simulator on x86_64


ErikEverson:
tvOS support is important for us as well so we don't have to maintain a fork as well. 
cc: @phatblat

mrmonday:
Cool, thanks!

PR #662, opened 2023-11-20
Add ICMP Destination unreachable Next-hop MTU

### LINKED ISSUE #661: ICMP Destination unreachable Next-hop MTU is not exposed
According to https://en.wikipedia.org/wiki/Internet_Control_Message_Protocol#Destination_unreachable, the "Next-hop MTU" is part of the header, but in libpnet, https://docs.rs/pnet/latest/pnet/packet/icmp/destination_unreachable/index.html the entire header is marked as unused.

fabi321:
This commit adds the Next-hop MTU as described in https://en.wikipedia.org/wiki/Internet_Control_Message_Protocol#Destination_unreachable

Fixes #661


mrmonday:
Thanks!

PR #673, opened 2024-01-14
remove unnecessary mut statements

cppcoffee:
The `Sender::send` does not have to use the mut statement.


mrmonday:
This needs a little more explanation, since it's an API change. This also needs a rebase, please.

cppcoffee:
Sorry, I forgot the usage scenario I ran into. close PR.

PR #697, opened 2024-06-14
Fix unbound variable reference in build.sh

y-koj:
This commit fixes build.sh to run properly when environment variable $VERBOSE is not set.

Previously, `set -u` in build.sh caused the script to fail with `build.sh: line 170: $VERBOSE: unbound variable`
when environment variable $VERBOSE is not defined.
This fix introduces a check to set $VERBOSE to 0 if the variable is not set.

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -62,6 +62,12 @@ if [[ -z "$PNET_TEST_IFACE" && "$SYSTEM" = "Linux" ]]; then
     done
 fi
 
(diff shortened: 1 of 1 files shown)


hawkinsw:
I can confirm that I have tested this fix and it resolved the error. Thank you @y-koj !

mrmonday:
Thanks!

y-koj:
Thank you for the review and merge!

PR #701, opened 2024-06-15
Bugfix for DNS response (length calculation) and further decoding of RRs.

sharmadpk:
1. Fixed bugs in length calculation in fields of DNS responses - answer, authoritative and additional.
2. Added support to decode data for SRV services and put together any string (using C0 construct).
3. Tested against mDNS packets

### DIFF
--- a/pnet_packet/src/dns.rs
+++ b/pnet_packet/src/dns.rs
@@ -588,24 +588,39 @@ pub struct SrvName {
 impl SrvName {
     pub fn new(name: &str) -> Self {
         let parts: Vec<&str> = name.split('.').collect();
-        let (instance, service, protocol, domain) = match parts.as_slice() {
-            [instance, service, protocol, domain @ ..] if service.starts_with('_') && protocol.starts_with('_') => {
-                (Some(String::from(*instance)), Some(String::from(*service)), Some(String::from(*protocol)), Some(String::from(domain.join("."))))
+        match parts.len() {
+            len if len >4 => {
+                SrvName {
+                    instance: Some(parts[0..len-3].join(".")),
+                    service: Some(String::from(parts[len-3])),
+                    protocol: Some(String::from(parts[len-2])),
+                    domain: Some(String::from(parts[len-1])),
+                }
             },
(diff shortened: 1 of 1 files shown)


pnet_packet/src/dns.rs
```diff
@@ -467,6 +468,133 @@ pub struct DnsResponse {
...
+            name = start;
+            continue;
+        }
+        if !rname.is_empty() {
+            rname.push('.');
+        }
+        rname.push_str(
+            str::from_utf8(&name[offset + 1..offset + 1 + label_len as usize])
+                .ok()
+                .unwrap(),
```
mrmonday, line 519: I think this method should probably return a `Result` instead, rather than crashing on invalid utf8.

pnet_packet/src/dns.rs
```diff
@@ -467,6 +468,133 @@ pub struct DnsResponse {
...
+        }
+        size += label_len + 1;
+        offset += label_len + 1;
+    }
+    size
+}
+
+// parse any string in the DNS message, using its unique encoding
+// Call this on DnsPacket, pass the string([u8]) to be decoded
+pub fn dns_parse_name(packet: &DnsPacket, coded_name: &Vec<u8>) -> String {
```
mrmonday, line 492: Probably just `parse_name()`? No need for the prefix.

mrmonday:
There's quite a lot of logic in here, and I don't know anyway near enough about DNS to review this properly.

I'd appreciate it if someone else could take a look over it before merging.

pnet_packet/src/dns.rs
sharmadpk: Accepted. I overlooked this while copying the code from the qname funciton. I will make this change here and the corresponding test cases. Todo: we should use the punycode crate for the real validity check and return the true string. utf8 is not allowed in the DNS messages anyway.

pnet_packet/src/dns.rs
sharmadpk: Accepted.

sharmadpk: done

PR #702, opened 2024-06-17
Add (Linux) support for configuring path MTU discovery

hawkinsw:
On Linux, setting the socket's path MTU discovery option is the way to set each packet's Don't Fragment bit, albeit indirectly.


hawkinsw:
As I said before, thank you *all* for the work that you do maintaining this awesome library. I hope that this added functionality is helpful!

pnet_transport/src/lib.rs
```diff
@@ -197,12 +197,12 @@ pub fn transport_channel_with(
...
-fn set_sockopt_u8(
+/// Sets a socket option whose value is a signed 32-bit integer. Close the socket on error.
+fn set_sockopt_i32(
     socket: Arc<pnet_sys::FileDesc>,
     level: libc::c_int,
     name: libc::c_int,
-    value: u8,
+    value: i32,
 ) -> io::Result<()> {
     let value = value as i32;
```
mrmonday, line 207: This line is no longer needed, I think?

pnet_transport/src/lib.rs
```diff
@@ -269,6 +281,19 @@ impl TransportSender {
         set_sockopt_u8(self.socket.clone(), level, name, tos as u8)
     }
 
+    /// Configures path MTU discovery on a socket.
+    ///
+    /// On Linux-based systems, configuring the socket's path MTU discovery setting is
+    /// the way to "(un)set" the "don't fragment" bit in the IP header.
+    #[cfg(target_os = "linux")]
+    pub fn configure_path_mtu_discovery(&mut self, enable_disable: bool) -> io::Result<()> {
```
mrmonday, line 289: Is there a reason to name this differently to the other methods that set options?

pnet_transport/src/lib.rs
```diff
@@ -269,6 +281,19 @@ impl TransportSender {
...
+        let (level, name, value) = match self.channel_type {
+            Layer4(Ipv4(_)) | Layer3(_) => (pnet_sys::IPPROTO_IP, pnet_sys::IP_MTU_DISCOVER, if enable_disable { pnet_sys::IP_PMTUDISC_DO } else {pnet_sys::IP_PMTUDISC_DONT  }),
+            Layer4(Ipv6(_)) => (pnet_sys::IPPROTO_IPV6, pnet_sys::IPV6_MTU_DISCOVER, if enable_disable { pnet_sys::IPV6_PMTUDISC_DO } else {pnet_sys::IPV6_PMTUDISC_DONT  }),
+        };
```
mrmonday, line 293: This match statement is a bit intense :wink: 

Please reformat or refactor this so it's easier to read.

pnet_transport/src/lib.rs
hawkinsw: Yes! Sorry I missed that!

pnet_transport/src/lib.rs
hawkinsw: Absolutely! I think that I got a little carried away!!

pnet_transport/src/lib.rs
hawkinsw: I thought about this a ton and thought of two reasons:
1. It's how the option is "labeled" in Linux
2. The knock-on effect of changing the IP header option is just that: a knock-on effect.

That said, I agree with your sentiment because I added the option *just* so that I could set that bit in the IP header.

I will ultimately defer to you as the expert!!

pnet_transport/src/lib.rs
hawkinsw: Fixed!

pnet_transport/src/lib.rs
hawkinsw: I have made a two-dimensional array that holds the different `level`, `name`, `value` combinations and use the `self.channel_type` and `enable_disable` to generate indexes. This seemed to me like a reasonable tradeoff between making that monstrous match (alliteration!) more readable and minimizing code duplication. If you still think that it is too unreadable, please let me know -- I'd be glad to take another swing at it.

PR #703, opened 2024-06-20
Enhance EtherType display to show name and hexadecimal value

Akmot9:
Now, when the EtherType number is not recognized, it displays the hexadecimal value for further analysis.


mrmonday:
Looks good, thanks!

PR #704, opened 2024-06-21
fix: if bpfN is less than config.bpf_fd_attempts, it will cause Err to be ENOENT.

shenjinti:
In MacOS`/dev/bpf[0-4]`, the default value of `bpf_fd_attempts` is `1000`

It will cause the Err of get_fd to be **ENOENT**, which is not a correct error. 

This patch records the last errno, such as PermissionDenied.


mrmonday:
Looks good, thanks!

PR #562, opened 2022-04-28
enable #[no_std] use by providing std feature

Felix-El:
Also adds a dependency on no_std_net for no_std addresses.

https://github.com/libpnet/libpnet/issues/133

PR #712, opened 2024-07-09
Use `core::net` instead of depending on `no-std-net`

faern:
As I wrote in a comment a few months ago (https://github.com/libpnet/libpnet/issues/133#issuecomment-2071615766). `core::net` is stable since Rust 1.77. So we can get rid of `no-std-net` here.

According to [this comment](https://github.com/libpnet/libpnet/pull/562#issuecomment-1136129586), it's fine to not support very old Rust versions. And I support that. 1.77 is soon three versions behind already, and not having to use custom IP addr types is really nice.

### DIFF
--- a/pnet_base/Cargo.toml
+++ b/pnet_base/Cargo.toml
@@ -12,14 +12,13 @@ categories = ["network-programming"]
 edition = "2021"
 
 [dependencies]
-no-std-net = { version = "0.6.0", default-features = false }
 serde = { version = "1.0.171", optional = true, default-features = false }
 
 [dev-dependencies]
 serde_test = "1.0.171"
 
 [features]
-std = ["no-std-net/std"]
+std = []
 default = ["std"]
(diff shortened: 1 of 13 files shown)


faern:
Ping @mrmonday, what do you think?

Given [this comment](https://github.com/libpnet/libpnet/pull/562#issuecomment-1136129586) I'll go ahead and merge this after some further time unless feedback is provided.

mrmonday:
I think this is good - we've had issues in the past with breaking std or no-std builds when changing these things, but I don't see anything here that would affect that.

mrmonday:
Re: versioning - I bump everything every release, whether there have been changes or not. This is mostly a convenience thing, since it saves tracking what has/hasn't changed, but it's also good from a testing/support perspective - no need to test all the different combinations, just "version X of pnet_* works with version X of pnet_*".

Issue #278, opened 2017-07-25
HELP 


vconstantin:
can someone help me, how to add this libpnet? i had installed nighlty version rust. and what should i do next. where i have to place this repository. how to build this? Thanks

polachok:
What are you trying to achieve? I'd recommend reading the rust book first: https://doc.rust-lang.org/book/second-edition/

mrmonday:
If you want to include libpnet in your project, you can add it as a dependency: https://github.com/libpnet/libpnet#usage

If those instructions don't make sense, I'd advise looking through the book that @polachok shared.

If there's something more specific, I'm sure we can help further :)

vconstantin:
@polachok @mrmonday thankyou. I have 2 more questions. 
1. where's the root of this repository to place packet.lib from WinPcap?
2. i got this error "linking with `gcc` failed: exit code: 1" when cargo test

im using Rust nightly 1.20 64bit on windows 10. 
im new for this thing so thank you for your help

mrmonday:
I don't have a Windows machine to test with right now, but try this:

1. In your `Cargo.toml`, in the `[package]` section, make sure you have:

```toml
build = "build.rs"
```

2. Create a `build.rs` file, in the same directory as `Cargo.toml`:
```rust
fn main() {
    use std::env;

    let manifest_dir = env::var("CARGO_MANIFEST_DIR").unwrap();
    println!("cargo:rustc-link-search=native={}/lib", manifest_dir);
}
```
3. Create a `lib` directory in the root of your project (the same place as `Cargo.toml` and `build.rs`)
4. Download the WinPcap Developers Pack from https://www.winpcap.org/devel.htm, and extract the `WpdPack/Lib/x64/Packet.lib` file, placing it into the `lib` directory you just created.

I realise that these instructions aren't in the libpnet README - if this works for you let me know and I'll update the README (or you can make a pull request if you like :))

vconstantin:
pnettest::check_test_environment
    pnettest::layer2
    pnettest::layer3_ipv4
    pnettest::layer4_ipv4
    pnettest::layer4_ipv6

still got this problem. @mrmonday

mrmonday:
As stated in the error message, you need to set the environment variable `RUST_TEST_THREADS=1` before running the tests.

Sorry for the slow response.

mrmonday:
Is this resolved now?

mrmonday:
[...]

PR #714, opened 2024-07-25
docs: update Windows build instructions for linking WinPcap library

xairaven:
This pull request updates the README to provide clearer instructions on how to link the WinPcap library to the project on Windows. The changes include an additional method for resolving common integration issues by using a `build.rs` script to correctly configure the library path.

I struggled with the `linking with x86_64-w64-mingw32-gccfailed: exit code 1` error. Then found a solution in the issues here: https://github.com/libpnet/libpnet/issues/278#issuecomment-318155351


xairaven:
Additional note: This clarification in the README.md is not intended to build the library project, but to build the project that uses the crate. This README is displayed on the [crate's page](https://crates.io/crates/pnet) on crates.io. Accordingly, I tried to build my project following those instructions, but it didn't work. Perhaps the phrase "put Packet.lib in a directory named lib, in the root of this repository" will confuse the user, but I think they will understand that it applies not only to the library project but also to the local project where the crate is used.

PR #717, opened 2024-10-03
Draft vlan support

Belpaire:
This is a draft for the ability to add vlan packet sniffing to raw linux sockets.  At the moment vlan tags are automatically stripped by the linux kernel and one needs some auxiliary constructs to reconstitute the vlan packet as it arrived on the wire.

Note that to demonstrate this ability I added an extra method to datalinkreceiver called next_msg, this is to demonstrate to how accessing the data construct for reconstituting vlan traffic is done, I don't want to add this method in the final version, it's for demo purposes (so you can see the differences in the existing next function like using recvmsg instead of recvfrom). In the final version we'd use the info of the returned tpacket_auxdata struct to add the vlan tag back in the bytes we return in the 'next' method.

The main gist of reconstituting vlan traffic is by enabling PACKET_AUXDATA as a socket option, when using the recv_msg syscall the kernel will then also return a control message, the struct is specified in https://man7.org/linux/man-pages/man7/packet.7.html as well as explanation of the PACKET_AUXDATA value, I also copied the specification of the struct as a comment in this draft for clarification (which I will remove from the final commit likely).

Mainly I'd like some comments on how best to approach adding the vlan traffic sniffing. I think the most straight forward way would be to enable an option in the config struct to enable reconstituting the vlan traffic as a config option (with it defaulting to false, so we don't change any end users relying on the kernel already stripping vlan tags), in which case we will enable PACKET_AUXDATA and if we detect if it is truly a VLAN packet to add the bytes back as to how they were transmitted on the wire. 

We can default to using recv_msg instead of recv_from in the next method as recv_msg is basically a more flexible variant of all the recv family of syscalls.

Please let me know your thoughts.

### DIFF
--- a/pnet_datalink/src/bindings/linux.rs
+++ b/pnet_datalink/src/bindings/linux.rs
@@ -37,3 +37,25 @@ pub struct packet_mreq {
     pub mr_alen: libc::c_ushort,
     pub mr_address: [libc::c_uchar; 8],
 }
+
+#[repr(C)]
+pub struct tpacket_auxdata {
+    pub tp_status: libc::c_uint,
+    pub tp_len: libc::c_uint,
+    pub tp_snaplen: libc::c_uint,
+    pub tp_mac: libc::c_ushort,
+    pub tp_net: libc::c_ushort,
+    pub tp_vlan_tci: libc::c_ushort,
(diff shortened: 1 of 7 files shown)


fhhowdy:
This 100% needs to be perfected and included in pnet.

PR #720, opened 2024-10-09
Fix cfgs to fix build with rust 1.80

joshg4096:
* With rust 1.80, cfgs are checked to be valid. See https://blog.rust-lang.org/2024/05/06/check-cfg.html
  * Remove deprecated clippy features: https://blog.rust-lang.org/2024/02/28/Clippy-deprecating-feature-cargo-clippy.html
  * Add "nightly" feature
* The use of `#[cfg(all(test, feature = "benchmark"))]` was invalid as benchmark isn't a feature. This was causing the benchmarks to never be run. Move these to the correct directory for benchmarks to allow these to run. This was tested by running `rustup run nightly cargo bench`

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -21,6 +21,7 @@ travis = []
 serde = ["pnet_base/serde", "pnet_datalink?/serde"]
 std = ["pnet_base/std", "pnet_sys", "pnet_datalink", "pnet_transport", "ipnetwork"]
 default = ["std"]
+nightly = []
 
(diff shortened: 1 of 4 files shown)


hawkinsw:
I can confirm that this patch fixes build errors on `x86_64-redhat-linux` with rustc 1.85.0-nightly (c44b3d50f 2024-12-03). Thank you, @joshg4096 for submitting!

